PR #32, opened 2017-10-19
Added Decay support for bxt_autostop

Shigbeard:
According to the leaderboard, time occurs on the first frame of the alienflyer's explosion. On death, monster_alienflyer triggers previctory_mm. Extensive testing and map extracting has confirmed that this is the only instance where previctory_mm is triggered, with no unintended side effects.


YaLTeR:
Thanks!

PR #36, opened 2018-04-15
Display absolute health

philippTheCat:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -2444,7 +2444,7 @@ void HwDLL::SetPlayerVelocity(float velocity[3])
 	player.Velocity[2] = velocity[2];
 }
 
-bool HwDLL::TryGettingAccurateInfo(float origin[3], float velocity[3])
+bool HwDLL::TryGettingAccurateInfo(float origin[3], float velocity[3], float* health)
```
mikesmiffy128: I'm no C++ expert but isn't generally advised to use a reference instead of a pointer for out parameters?

BunnymodXT/modules/HwDLL.hpp
```diff
@@ -130,7 +130,7 @@ class HwDLL : public IHookableNameFilterOrdered
 
 	void SetPlayerOrigin(float origin[3]);
 	void SetPlayerVelocity(float velocity[3]);
-	bool TryGettingAccurateInfo(float origin[3], float velocity[3]);
+	bool TryGettingAccurateInfo(float origin[3], float velocity[3],float* health);
```
mikesmiffy128: missing a space before "float"

BunnymodXT/hud_custom.cpp
```diff
@@ -643,6 +643,16 @@ namespace CustomHud
 		}
 	}
 
+	static void DrawHealth(float flTime)
```
YaLTeR: The rest of the code doesn't use `static`, be consistent with the rest of the code.

BunnymodXT/hud_custom.cpp
```diff
@@ -643,6 +643,16 @@ namespace CustomHud
...
+	{
+		if (CVars::bxt_hud_healthmeter.GetBool())
+		{
+			int x, y;
+			GetPosition(CVars::bxt_hud_healthmeter_offset, CVars::bxt_hud_healthmeter_anchor, &x, &y, 0, -4 * NumberHeight);
+			DrawNumber((int) player.health, x, y);
```
YaLTeR: `static_cast<int>(player.health)`

BunnymodXT/modules/HwDLL.cpp
mikesmiffy128: Yeah I personally don't like the way references look but I figured it was worth suggesting anyway.

PR #39, opened 2018-05-03
Add TriAPI trigger rendering and wallhack

YaLTeR:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -2901,3 +2979,46 @@ HOOK_DEF_0(HwDLL, void, __cdecl, Cmd_Exec_f)
...
+// we want NPCs to remain opaque, to make them more visible. This function draws
+// the worldspawn and other brush entities but not studio models (NPCs).
+HOOK_DEF_2(HwDLL, void, __cdecl, R_DrawSequentialPoly, msurface_t *, surf, int, face)
+{
+	if (!(CmdFuncs::IsCheating() && CVars::bxt_wallhack.GetBool())) {
+		ORIG_R_DrawSequentialPoly(surf, face);
+		return;
+	}
+
+	glEnable(GL_BLEND);
```
YaLTeR: What's gonna happen on D3D and Software?

BunnymodXT/modules/ServerDLL.hpp
```diff
@@ -51,8 +51,13 @@ class ServerDLL : public IHookableDirFilter
 	std::vector<const edict_t *> GetUseableEntities(const Vector &origin, float radius) const;
 	std::vector<const Vector *> GetNodePositions() const;
 
+	inline const char *GetString(int string) const {
+		if (!ppGlobals)
+			return nullptr;
```
YaLTeR: Other functions you added don't check for the pointers being found. Maybe drop the check here?

BunnymodXT/triangle_drawing.cpp
```diff
@@ -130,7 +130,120 @@ namespace TriangleDrawing
 		}
 	}
 
+	static void GetTriggerColor(const char *classname, bool inactive, float *r, float *g, float *b, float *a)
```
YaLTeR: Put this somewhere shared so there's no code duplication between this and the original spot?

PR #48, opened 2018-09-21
Add jumpbug autofunc

arianon:


YaLTeR:
Very nice! I'll test this later today, LGTM otherwise. @Matherunner please take a look too.

YaLTeR:
Tested; seems to work well.

arianon:
Something that throws me off about using `HLStrafe::GetPositionType` is [this `player.Velocity[2]` check](https://github.com/HLTAS/hlstrafe/blob/18412a7cf545f3feb0b3c8d3342fd898a31e293f/src/hlstrafe.cpp#L855-L856), which should be always true if we're falling, even though the autojumpbug works fine. An alternative implementation for checking whether to do the JB is this

```c++
float nextOrigin[3];
for (std::size_t i = 0; i < 3; ++i)
	nextOrigin[i] = player.Origin[i];

// Check if there's ground 2 units below us.
nextOrigin[2] -= 2;
auto tr = PlayerTrace(player.Origin, nextOrigin, HLStrafe::HullType::DUCKED);
if (!tr.StartSolid && !tr.AllSolid) {
	// Jumpbug if we're about to land
	Jump = true;
	Duck = false;
}
```

But I am not completely sure what `PlayerTrace.StartSolid` and `PlayerTrace.AllSolid` are supposed to be, so I'm not confident on this implementation.

Tomorrow I'll take a closer, more thorough look at my code.

YaLTeR:
When the player is falling their vertical speed is negative, and that check is for positive speeds above 180.

PR #49, opened 2018-10-04
Add collision hull drawing

Matherunner:
Add bxt_collision_hulls and bxt_collision_hulls_alpha.

There's still some problems with this, specifically with the hulls of brush models:

- [ ] collision planes tend to totally obscure parts of them even with transparency
- [ ] the shapes of the hulls of brush models do not always represent their real collision
- [ ] hulls of brush models may disappear if the actual entity itself is not in view

These may or may not be deal breaking problems.

### DIFF
--- a/BunnymodXT/draw_hulls.cpp
+++ b/BunnymodXT/draw_hulls.cpp
@@ -1,0 +1,912 @@
+/*
+Copyright (C) 2018 Matherunner
+Copyright (C) 2016 Uncle Mike
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+#include "draw_hulls.hpp"
+#include "hlsdk/common/const.h"
+#include <GL/gl.h>
+
+namespace DrawHulls
+{
+	struct rgb {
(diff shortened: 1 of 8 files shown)


BunnymodXT/draw_hulls.cpp
```diff
@@ -0,0 +1,912 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+#include "draw_hulls.hpp"
+#include "hlsdk/common/const.h"
```
arianon: Did you mean to write `HLSDK/common/const.h`? paths are case-sensitive on Linux so I had to change this to compile this file.

Matherunner: I’ve only tested in this the windows vm. I’ll fix this.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3112,3 +3191,43 @@ HOOK_DEF_2(HwDLL, byte *, __cdecl, Mod_LeafPVS, mleaf_t *, leaf, model_t *, mode
...
+	ORIG_R_DrawBrushModel(ent);
+	hullInfo.set_in_draw_brush(false);
+}
+
+HOOK_DEF_2(HwDLL, void, __cdecl, R_RotateForEntity, float *, origin, cl_entity_s *, ent)
+{
+	ORIG_R_RotateForEntity(origin, ent);
+
+	if (CVars::bxt_collision_hulls.GetInt()) {
+		constexpr ptrdiff_t offset_model = 0xb94;
```
YaLTeR: Hmm, is this the same across the engines and OSes and whatnot?

Issue #50, opened 2018-12-31
[Enhancement] `R_DrawSequentialPoly` pattern for Opposing Force


quantumdude836:
Similar to "HL-4554" pattern (and same length), but the prologue is slightly different. This enables `bxt_wallhack` in OpForce.

`A1 ?? ?? ?? ?? 53 55 BD 01 00 00 00 8B 88 F8 02 00 00 56 3B CD 57 75 62 E8 ?? ?? ?? ?? 68 03 03 00 00 68 02 03 00 00`

YaLTeR:
Thanks, could you submit a pull request?

PR #51, opened 2018-12-31
Add R_DrawSequentialPoly pattern for Opposing Force

quantumdude836:
Pull request for #50


YaLTeR:
Yeah, there's a number of versions, that's why I'm asking. I'd prefer the pattern name to say the engine version too. Could you change it to `OpposingForce-1600`?

YaLTeR:
Thanks!

PR #89, opened 2020-01-23
Add bxt_collision_depth_map_pixel_scale for rendering the collision map at lower resolutions

Marckvdv:
Right now, using the collision depth map of BXT is very slow at high resolutions. This PR adds the console variable `bxt_collision_depth_map_pixel_scale` which indicates the size of each traced pixel (i.e. a pixel scale of 4 would mean that a single ray is cast for a 4x4 area, instead of the old 1 ray per pixel). 

PR also includes a rewrite of the code that calculates the camera rays, which is both simpler and fixes the old depth map not completely lining up with the normal HL render. The view was especially warped when looking at walls at a shallow angle, which is now fixed.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -879,9 +879,11 @@ namespace CustomHud
 			glTexEnvf(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_MODULATE);
 			glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
+			const auto pixel_scale = CVars::bxt_collision_depth_map_pixel_scale.GetInt();
+
 			// Main loop.
-			for (int y = 0; y < si.iHeight; ++y) {
-				for (int x = 0; x < si.iWidth; ++x) {
+			for (int y = 0; y < si.iHeight; y += pixel_scale) {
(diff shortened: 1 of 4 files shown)


BunnymodXT/hud_custom.cpp
```diff
@@ -950,24 +928,26 @@ namespace CustomHud
...
-					glVertex2i(x + 1, y + 1);
-					glVertex2i(x + 1, y);
+					glVertex2i(x, y + pixel_scale);
+					glVertex2i(x + pixel_scale, y + pixel_scale);
+					glVertex2i(x + pixel_scale, y);
 					glEnd();
 				}
 			}
 
+			EngineDevMsg("Performed collision draw");
```
YaLTeR: Also there's a left-over debug print.

Marckvdv:
Took a while but should be fine now. Split into two commits, removed the debug line, fixed warnings.

YaLTeR:
> Took a while but should be fine now. Split into two commits, removed the debug line, fixed warnings. Or did you want the ray calculations in a separate PR?

No, thanks, this is perfect. I will test it now.

YaLTeR:
This is really cool! Works very well, and the artifacts are gone too. Thanks a lot!

I think I'll default the pixel scale to 8 so people can enable the collision depth map on their regular resolution safely.

PR #93, opened 2020-02-01
bxt_hud_tas_editor_status shows frame bulk pitch and yaw

Eddio0141:


BunnymodXT/hud_custom.cpp
```diff
@@ -1254,5 +1266,10 @@ namespace CustomHud
 		frame_bulk_status.attack1 = frame_bulk.Attack1;
 		frame_bulk_status.attack2 = frame_bulk.Attack2;
 		frame_bulk_status.reload = frame_bulk.Reload;
+
+		frame_bulk_status.pitch = frame_bulk.GetPitch();
+		frame_bulk_status.yaw = frame_bulk.GetYaw();
```
YaLTeR: These two lines should only be called when their corresponding angle is present, so
```cpp
if (frame_bulk.GetYawPresent()) {
    frame_bulk_status.yaw_present = true;
    frame_bulk_status.yaw = frame_bulk.GetYaw();
}
```
otherwise it can hit an assert in `GetPitch` and `GetYaw`.

BunnymodXT/hud_custom.cpp
```diff
@@ -1254,5 +1266,15 @@ namespace CustomHud
 		frame_bulk_status.attack1 = frame_bulk.Attack1;
 		frame_bulk_status.attack2 = frame_bulk.Attack2;
 		frame_bulk_status.reload = frame_bulk.Reload;
+
+		frame_bulk_status.pitch_present = frame_bulk.PitchPresent;
+		if (frame_bulk_status.pitch_present) {
+			frame_bulk_status.pitch = frame_bulk.GetYaw();
```
YaLTeR: A copy-paste error: should be `GetPitch()`.

YaLTeR:
Thanks!

PR #96, opened 2020-03-15
Add +/-bxt_triggers_place

Marckvdv:
# Usage
This PR add +bxt_triggers_place and -bxt_triggers_place. These allow easier placement of custom triggers by simply looking at the begin point and end point of the trigger to be created (+bxt_triggers_place indicates the starting point of the trigger and -bxt_triggers_place indicates the end point). 

Video demonstration: https://streamable.com/nnrwl

# Code
The code isn't great at the moment, adding two globals to the `custom_triggers.hpp` file. This is somewhat due to the already present structure of maintaining the custom triggers. In case the code is not fine like this, I was working on a small rewrite of the custom triggers in my custom_triggers_rewrite branch which I could append to this PR when it is done.

### DIFF
--- a/BunnymodXT/custom_triggers.cpp
+++ b/BunnymodXT/custom_triggers.cpp
@@ -7,6 +7,8 @@
 namespace CustomTriggers
 {
 	std::vector<Trigger> triggers;
+	Vector place_start;
+	bool placing = false;
 
 	void Trigger::normalize()
 	{
@@ -72,6 +74,13 @@ namespace CustomTriggers
 			if (last_char != '\n' && last_char != ';')
 				command += '\n';
 		}
+	}
+
+	void Trigger::set_corner_positions(Vector corner1, Vector corner2)
+	{
+		corner_min = corner1;
+		corner_max = corner2;
+		normalize();
 	}
 
 	void Trigger::update(const Vector& player_position, bool ducking)
@@ -171,6 +180,20 @@ namespace CustomTriggers
 	{
 		for (auto& trigger : triggers)
 			trigger.update(player_position, ducking);
+
+		if (placing) {
+			if (!triggers.empty())
+			{
+				auto trace = HwDLL::GetInstance().CameraTrace();
+				Vector place_end(trace.EndPos);
+
+				triggers.back().set_corner_positions(place_start, place_end);
+			}
+			else
+			{
+				placing = false;
+			}
+		}
 	}
 
 	void Update(const Vector& player_position_start, const Vector& player_position_end, bool ducking)
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -1879,6 +1879,41 @@ struct HwDLL::Cmd_BXT_Triggers_Add
 	}
 };
 
+struct HwDLL::Cmd_BXT_Triggers_Place_Down
+{
+	NO_USAGE();
+
+	static void handler()
+	{
+		auto trace = HwDLL::GetInstance().CameraTrace();
+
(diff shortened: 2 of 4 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -2620,6 +2655,8 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 		Cmd_BXT_Triggers_SetCommand,
 		Handler<const char*>,
 		Handler<unsigned long, const char*>>("bxt_triggers_setcommand");
+	wrapper::Add<Cmd_BXT_Triggers_Place_Down, Handler<>, Handler<const char*>>("+bxt_triggers_place");
+	wrapper::Add<Cmd_BXT_Triggers_Place_Up, Handler<>, Handler<const char*>>("-bxt_triggers_place");
```
YaLTeR: Please leave only the `void` handler as the `const char*` one is unused. Remove the functions too.

BunnymodXT/modules/HwDLL.hpp
```diff
@@ -199,6 +199,7 @@ class HwDLL : public IHookableNameFilterOrdered
 	}
 
 	HLStrafe::TraceResult PlayerTrace(const float start[3], const float end[3], HLStrafe::HullType hull, bool extendDistanceLimit = false);
+	HLStrafe::TraceResult CameraTrace(float max_distance = 1000.f);
```
YaLTeR: I think I usually used 8192 for this distance.
```suggestion
	HLStrafe::TraceResult CameraTrace(float max_distance = 8192);
```

BunnymodXT/custom_triggers.cpp
```diff
@@ -7,6 +7,8 @@
 namespace CustomTriggers
 {
 	std::vector<Trigger> triggers;
+	Vector place_start;
```
YaLTeR: I don't think this is actually needed. Just take the first corner of the last trigger when placing.

BunnymodXT/modules/HwDLL.cpp
Marckvdv: That is what I initially had, but without the `const char*` handlers, binding the commands to a key does not work properly. This is probably the same reason that commands like `+bxt_tas_ducktap`, `+bxt_tas_autojump` have `const char*` handlers though they don't use them.

BunnymodXT/custom_triggers.cpp
Marckvdv: I thought about removing this, but I think this breaks due to the normalization of the trigger corners. Because the endpoint corner can change position, it could be that the start point and the end point change between the Triggers `corner_min` and `corner_max`.

BunnymodXT/modules/HwDLL.cpp
YaLTeR: Oh, right, that's because the engine passes in a button code. Okay then.

BunnymodXT/custom_triggers.cpp
YaLTeR: Ah, that's right.

YaLTeR: Could you note that in the comment there?

Marckvdv: I've added a comment in the header explaining this. Combined all the changes into a single commit.

PS: Sorry if I'm doing this wrong, I'm not sure how I was meant to change this. If used several commits you could've more easily seen what I changed but then it becomes a mess in the commit history 🤷‍♂

YaLTeR:
Thanks!

PR #105, opened 2020-06-06
ci: Minor changes

Margen67:
Disable fail-fast.
Formatting.
Properly checkout submodules.
Add -j to build.
Use runner.os in artifact name.
Make mkdir verbose.

Windows:
 Use env var for rapidjson version, use curl to download rapidjson.
 Change / to \.


YaLTeR:
Thanks, but I changed some things back according to what I like more.

PR #110, opened 2020-06-21
Save all cvars to demos

hobokenn:


YaLTeR:
Oh, no, it should just walk the cvars and save each of them. This way it'll be much faster and also have all custom mod cvars.

PR #124, opened 2020-08-05
Add PM_Jump pattern for Paranoia

mxpph:
Allows autojump.

### DIFF
--- a/BunnymodXT/patterns.hpp
+++ b/BunnymodXT/patterns.hpp
@@ -398,9 +398,7 @@ namespace patterns
 			"AG-Client",
 			"51 8B 0D ?? ?? ?? ?? 8B 81 D0 00 00 00 85 C0 74 09 83 89 C8 00 00 00 02 59 C3 D9 05 ?? ?? ?? ?? D9 81 CC 00 00 00 DA E9",
 			"BigLolly",
-			"55 8B EC 83 EC 14 C7 45 F4 00 00 00 00 C7 45 FC 00 00 00 00 A1 ?? ?? ?? ?? 83 B8 D0 00 00 00 00 74 1F 8B 0D ?? ?? ?? ?? 8B 91 C8 00 00 00 83 CA 02 A1",
-			"Paranoia",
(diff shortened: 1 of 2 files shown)


BunnymodXT/patterns.hpp
```diff
@@ -398,7 +398,9 @@ namespace patterns
 			"AG-Client",
 			"51 8B 0D ?? ?? ?? ?? 8B 81 D0 00 00 00 85 C0 74 09 83 89 C8 00 00 00 02 59 C3 D9 05 ?? ?? ?? ?? D9 81 CC 00 00 00 DA E9",
 			"BigLolly",
-			"55 8B EC 83 EC 14 C7 45 F4 00 00 00 00 C7 45 FC 00 00 00 00 A1 ?? ?? ?? ?? 83 B8 D0 00 00 00 00 74 1F 8B 0D ?? ?? ?? ?? 8B 91 C8 00 00 00 83 CA 02 A1"
+			"55 8B EC 83 EC 14 C7 45 F4 00 00 00 00 C7 45 FC 00 00 00 00 A1 ?? ?? ?? ?? 83 B8 D0 00 00 00 00 74 1F 8B 0D ?? ?? ?? ?? 8B 91 C8 00 00 00 83 CA 02 A1",
+			"Paranoia",
+			"55 8B EC 83 EC 1C C7 45 ?? 00 00 00 00 C7 45 ?? 00 00 00 00"
```
YaLTeR, line 403: I would suggest just adding `??`s to two bytes in the BigLolly pattern to cover this case, however judging from the other change Paranoia also has a different ppmove offset. Thus, I suggest extending this pattern so it looks more different from the BigLolly pattern (assuming it diverges a little more down the line).

mxpph: > (assuming it diverges a little more down the line).
I noticed that it in fact did not, which made me decide to try the fix of changing the two different bytes to `??`. I expected it to crash from the offset, but it in fact did not. I must have made a mistake when calculating the offset the first time, and taken it from somewhere later in the function. Anyway, updated in latest commit.

YaLTeR:
Thanks!

PR #132, opened 2020-10-16
Add sequence and frame info to nihilanth HUD

Matherunner:
c4a30005


YaLTeR:
Thanks, could you also bring the TAS editor status one line down?

image

Matherunner:
OK I've moved it down.

image

PR #139, opened 2020-11-21
Fixed incorrect position for bxt_hud_entity_info

SmileyAG:
Before: 

before

Now:

now


YaLTeR:
I see that I changed it to -250 when going from entity_hp to entity_info, but upon some testing now classnames and targetnames seem as though they should fit with -200.

PR #149, opened 2021-05-18
Added mouse 4 multi drag in TAS Editor

Eddio0141:
Pressing mouse 4 and dragging a frame bulk will also drag multiple frame bulks with the same angle as the selected one as long as mouse 4 is held down.


YaLTeR:
Looks great, thanks!

PR #150, opened 2021-05-19
Added bxt_ch_set_angles

### LINKED ISSUE #147: bxt_ch_set_angles <pitch> <yaw>

chinese-soup:
Closes #147 

EDIT: Just noticed that there's some floating point error in setting the yaw, not sure why, but `bxt_ch_set_angles 10 10` ends up as:

image

Any idea how to fix that?

`hudSetViewAngles` is:
```c++
void hudSetViewAngles(float *va)

{
  (*g_engdstAddrs.SetViewAngles)(&va);
  cl.viewangles[0] = (vec_t)*va;
  cl.viewangles[1] = (vec_t)va[1];
  cl.viewangles[2] = (vec_t)va[2];
  return;
}
```

Also, should the cheat also include the roll as an argument, seeing as it's a cheat anyway (and roll is sometimes set - e.g. when player dies and deathcam starts)? What do you think, @SmileyAG?

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -579,6 +579,12 @@ void HwDLL::FindStuff()
 		else
 			EngineDevWarning("[hw dll] Could not find hudGetViewAngles.\n");
 
+		ORIG_hudSetViewAngles = reinterpret_cast<_hudSetViewAngles>(MemUtils::GetSymbolAddress(m_Handle, "hudSetViewAngles"));
+		if (ORIG_hudSetViewAngles)
+			EngineDevMsg("[hw dll] Found hudSetViewAngles at %p.\n", ORIG_hudSetViewAngles);
+		else
+			EngineDevWarning("[hw dll] Could not find hudSetViewAngles.\n");
+
 		ORIG_SV_AddLinksToPM = reinterpret_cast<_SV_AddLinksToPM>(MemUtils::GetSymbolAddress(m_Handle, "SV_AddLinksToPM"));
 		if (ORIG_SV_AddLinksToPM)
 			EngineDevMsg("[hw dll] Found SV_AddLinksToPM at %p.\n", ORIG_SV_AddLinksToPM);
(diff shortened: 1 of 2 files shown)


SmileyAG:
Well done! 

If you wish add <roll> argument — you could do that, but make next thing:

For example, if i'll miss <roll> argument, could you set it by default to 0?

Because as far as I know, the roll angle is zero in most cases, but I could be wrong.

YaLTeR:
> there's some floating point error in setting the yaw

There's no error, it's anglemod. This is expected behavior.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -1755,6 +1761,22 @@ struct HwDLL::Cmd_BXT_CH_Set_Origin_Offset
 	}
 };
 
+
```
YaLTeR, line 1764: the usual
```suggestion
```

chinese-soup: :see_no_evil:

chinese-soup:
> > there's some floating point error in setting the yaw
> 
> There's no error, it's anglemod. This is expected behavior.

Alright, great.

> If you wish add "roll" argument — you could do that, but make next thing:
> 
> For example, if i'll miss "roll" argument, could you set it to 0 as like default?
> 
> Because as far as I know, the roll angle is zero in most cases.

Alright, done. :+1:

YaLTeR (approved):

chinese-soup:
Drafting to include Windows support, I guess.

chinese-soup:
@SmileyAG Works on Windows now as well, dunno since when since I've always force pushed like the smart person I am. ;)

YaLTeR:
> @YaLTeR Should I just remove the hudSetViewAngles hook that only works on Linux and use the ClientDLL exported engine functions for both Linux and Windows?

Makes sense. Also

image

chinese-soup:
Should be good to go now.

YaLTeR:
Could you also add a check for pEngfuncs being NULL? To do nothing in that case.

chinese-soup:
Done.

YaLTeR:
Thanks

Issue #75, opened 2020-01-06
TWHL Tower crashes on wine


YaLTeR:
When running on wine, TWHL Tower crashes with BXT on:
- `bxt_bhopcap 1` with `Z_Free: freed a pointer without ZONEID`,
- `map twhl-lobby;bxt_tas_loadscript twhl.hltas` silently.

No crashes happen on Windows. No `bxt_bhopcap` crashes happen if I remove the automatic setting of it to `0` in `ServerDLL.cpp`. Additionally, for some reason a single time it had no crashes under wine, but I could not reproduce that.

PR #160, opened 2021-07-03
Basic support for Half-Payne & Decay: Solo Mission

hobokenn:
No bhop cap removal for DSM, because it is inlined in PM_Jump similar to Linux but unlike on Linux i can replace one byte and it jumps over the code, but since bxt_bhopcap is registered in the engine and PM_PreventMegaBunnyJumping is never found, it crashes with `Z_Free: freed a pointer without ZONEID` when the cvar value is changed. I assume rolling back bxt_bhopcap registration to the server or registering some new arbitrary `bxt_bhopcap_server` for these types of new mods is not an option?

EDIT: Also note that Half-Payne's server dll will not hook out of the box because it's directory name is `dll` instead of `dlls`, but i don't consider it to be a major issue.


YaLTeR:
> since bxt_bhopcap is registered in the engine and PM_PreventMegaBunnyJumping is never found, it crashes with Z_Free: freed a pointer without ZONEID when the cvar value is changed

Huh, that's weird, maybe there's a missing check somewhere? I remember a similar crash in (twhl tower #75) and I believe I also had that crash during bxt-rs development when something about cvar string management was wrong.

hobokenn:
> 
> 
> > since bxt_bhopcap is registered in the engine and PM_PreventMegaBunnyJumping is never found, it crashes with Z_Free: freed a pointer without ZONEID when the cvar value is changed
> 
> Huh, that's weird, maybe there's a missing check somewhere? I remember a similar crash in (twhl tower #75) and I believe I also had that crash during bxt-rs development when something about cvar string management was wrong.

Remember someone launched HL-WON from the package, did `bxt_tas_new`, it inserted some bxt_bhopcap 0 defaults, then they launched the TAS and crashed, basically the same thing.

hobokenn:
apparently you can't do that https://github.com/YaLTeR/BunnymodXT/blob/master/BunnymodXT/modules/ServerDLL.cpp#L1045

hobokenn:
tested with a couple of other cvars, seems to crash consistently 
image

hobokenn:
@YaLTeR could you review changes?

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -567,6 +584,9 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, PM_Jump)
 				MemUtils::ReplaceBytes(reinterpret_cast<void*>(pPMJump + offBhopcap), 6, reinterpret_cast<const byte*>("\x90\x90\x90\x90\x90\x90"));
 	}
 
+	if (pBhopcapWindows)
+		MemUtils::ReplaceBytes(reinterpret_cast<void*>(pBhopcapWindows), 1, reinterpret_cast<const byte*>(CVars::bxt_bhopcap_prediction.GetBool() ? "\x83" : "\x82"));
```
YaLTeR, line 588: What's the original and replacement assembly?

YaLTeR:
> tested with a couple of other cvars, seems to crash consistently
> image

Can either get rid of those or figure out how to fix.

BunnymodXT/modules/ClientDLL.cpp
hobokenn: 0f 83 jnc and 0f 82 jc

hobokenn:
> 
> 
> > tested with a couple of other cvars, seems to crash consistently
> > image
> 
> Can either get rid of those or figure out how to fix.

not sure what do you mean, i didn't add new set calls and i figured out how to not crash (at least for this mod), you were right - it was a missing check before calling set, i just didn't immediately connect the dots in my head as to what was happening, turns out you can't use set after registering a cvar.

hobokenn:
> 
> 
> > tested with a couple of other cvars, seems to crash consistently
> > image
> 
> Can either get rid of those or figure out how to fix.

I think i see what you mean.. that would be really unfortunate though, for example, to verify WON runs one runner has bxt_bhopcap 1, another one 0, will have to update goldsrc packages to set it to 0 in userconfigs. Yeah but it should be better than outright crashing.

YaLTeR:
Sorry for the delays in responses. If we can't use `.Set()`, maybe we can just use the usual engine call for cvar setting, or even just insert a `bxt_bhopcap 0` command?

hobokenn:
> 
> 
> Sorry for the delays in responses. If we can't use `.Set()`, maybe we can just use the usual engine call for cvar setting, or even just insert a `bxt_bhopcap 0` command?

Inserting a line works, nice! About the engine call: if you mean `Cvar_DirectSet` i don't know if it would work, that was actually the function that called `Z_Free` in the call stack.

YaLTeR:
So DirectSet works fine without crashes?

Also you'll need to fix the conflicts.

hobokenn:
> 
> 
> So DirectSet works fine without crashes?

ye it does

hobokenn:
this should be better 😄

YaLTeR:
Thanks!

PR #161, opened 2021-07-03
Pattern & offsets updates

hobokenn:
I realized using CBasePlayer::TakeDamage and CPushable::Move was not a good idea, because these functions are too similar between different dlls, instead use ForceClientDllUpdate.


YaLTeR (approved):

YaLTeR:
Thanks

PR #163, opened 2021-07-08
Add _bxt_interprocess_stop

### LINKED ISSUE #144: some way to stop or disable bunnysplit
more often than not i find myself not wanting to deal with livesplit for a short period of time, e.g. in the middle of attempts i might want to practice, run ILs or time some strat and for that i need to quit livesplit. it would be better if i could shutdown the pipe for the time being and then reset it when i want it back. so i was thinking of just adding a similar struct as the reset one, but with just shutdown(). don't know if there is more clever way of doing that.

hobokenn:
Closes #144


YaLTeR:
Thanks

PR #166, opened 2021-07-21
Log sv.max_edicts value to demos and fix typos in ServerDLL

hobokenn:


BunnymodXT/runtime_data.hpp
```diff
@@ -48,16 +48,21 @@ namespace RuntimeData
...
-	                            CVarValues,
-	                            Time,
-	                            BoundCommand,
-	                            AliasExpansion,
-	                            ScriptExecution,
-	                            CommandExecution,
-	                            GameEndMarker,
-	                            LoadedModules,
-				    CustomTriggerCommand>;
+								CVarValues,
```
YaLTeR, line 56: no, alignment is done using spaces, that way it doesn't break if you change tab size (as happened here):

image

hobokenn: like this?

YaLTeR: Yeah

hobokenn: cool, ty 👍

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4067,6 +4070,10 @@ void HwDLL::SaveInitialDataToDemo()
 
 	RuntimeData::Add(std::move(cvar_values));
 
+	const int maxEdicts = *reinterpret_cast<int *>(reinterpret_cast<uintptr_t>(sv) + offMaxEdicts);
```
YaLTeR, line 4073: Probably shouldn't do that if either the pointer or the offset is missing.

YaLTeR:
Thanks!

PR #168, opened 2021-07-24
Added bxt_disable_changelevel

SmileyAG:
This works on Windows, but I have no idea about Linux, can someone confirm it? @chinese-soup

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -1118,7 +1118,7 @@ void ServerDLL::RegisterCVarsAndCommands()
 	}
 	if (ORIG_CTriggerSave__SaveTouch)
 		REG(bxt_disable_autosave);
-	if (ORIG_CChangeLevel__ChangeLevelNow)
+	if (ORIG_CChangeLevel__ChangeLevelNow || ORIG_CChangeLevel__ChangeLevelNow_Linux)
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1092,6 +1117,8 @@ void ServerDLL::RegisterCVarsAndCommands()
 	}
 	if (ORIG_CTriggerSave__SaveTouch)
 		REG(bxt_disable_autosave);
+	if (ORIG_CChangeLevel__ChangeLevelNow)
```
hobokenn, line 1120: if (ORIG_CChangeLevel__ChangeLevelNow || ORIG_CChangeLevel__ChangeLevelNow_Linux)

hobokenn (changes requested):

BunnymodXT/modules/ServerDLL.cpp
hobokenn: could you add that to SaveTouch too please?

SmileyAG: Done. Thanks!

YaLTeR:
On Linux the function seems to be found but bxt_disable_changelevel 1 doesn't seem to do anything.

YaLTeR:
Thanks

PR #169, opened 2021-07-25
Patterns update for 4554 & NGHL

hobokenn:
Additionally added a couple of instructions for `V_FadeAlpha` just to avoid potential future collisions with SteamPipe and renamed HL-WON to HL-NGHL for `R_Clear` for consistency.


YaLTeR:
Thanks

PR #171, opened 2021-08-02
Added bxt_disable_bshift_outro_save

chinese-soup:
Quality of life feature for not accidentally crashing your game at the end of a run by saving the game.
Maybe needs some more checks that I don't know about... dunno + maybe find a way to only register the cvar (and maybe also only hook) when the game is bshift? 
Also missing Win patterns (like always).

https://github.com/ValveSoftware/halflife/issues/1014

### DIFF
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -43,6 +43,7 @@ namespace CVars
 	CVarWrapper bxt_show_triggers_legacy("bxt_show_triggers_legacy", "0");
 	CVarWrapper bxt_show_pickup_bbox("bxt_show_pickup_bbox", "0");
 	CVarWrapper bxt_disable_autosave("bxt_disable_autosave", "0");
+	CVarWrapper bxt_disable_bshift_outro_save("bxt_disable_bshift_outro_save", "0");
 	CVarWrapper bxt_disable_changelevel("bxt_disable_changelevel", "0");
 
 	// Clientside CVars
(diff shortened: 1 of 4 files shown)


YaLTeR (approved):

hobokenn:
I don't see the point of this, don't you need to save/load anyway to save time there? Or is this for casual players?

chinese-soup:
Sry didn't know about new strats.

> don't you need to save/load anyway to save time there? 
Yes, but I personally rather trade those 20 seconds for not accidentally crashing the game 60 % of the time at the end, when I don't have a time that has to compete with people using the save & load strat. 

But whatever, I'll just stop pressing my save key then.

hobokenn:
> Yes, but I personally rather trade those 20 seconds for not accidentally crashing the game 60 % of the time at the end, when I don't have a time that has to compete with people using the save & load strat.
> 
> But whatever, I'll just stop pressing my save key then.

yea i mean i don't want to argue, but we just had a discussion among mods about implementing something that would fix a crash in op4, which is way way way worse than this one, and we decided not to do it. i don't see how this fix is different.

chinese-soup:
> but we just had a discussion among mods about implementing something that would fix a crash in op4, which is way way way worse than this one, and we decided not to do it. i don't see how this fix is different.

Yeah, that's fine, I understand, I literally just didn't know about this strat. :smile: Although, the difference here is that it just disables saving on the level (at first I thought of just fixing the crash but then I realized that'd be cheaty), making it impossible to do strat, so it doesn't give you an advantage, just a 20 second disadvantage. 

> yea i mean i don't want to argue
No worries, I've closed it, now that I know it's meaningless since this strat is used and since I don't wanna pollute the code with this useless thing. :smile:

PR #172, opened 2021-08-03
Minor patches

hobokenn:


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -572,14 +572,9 @@ void ServerDLL::FindStuff()
...
 		if (pCZDS_Velocity_Byte) {
-			if (pattern == patterns::server::CZDS_Velocity_Byte.cend())
-				EngineDevMsg("[server dll] Found CZDS Velocity Reset Byte at %p.\n", pCZDS_Velocity_Byte);
-			else {
-				EngineDevMsg("[server dll] Found CZDS Velocity Reset Byte at %p (using the %s pattern).\n", pCZDS_Velocity_Byte, pattern->name());
-				offFuncObjectCaps = 0x18;
-			}
-		}
-		else {
+			EngineDevMsg("[server dll] Found CZDS Velocity Reset Byte at %p (using the %s pattern).\n", pCZDS_Velocity_Byte, pattern->name());
```
YaLTeR, line 575: Can pattern ever be == cend? Because then this will crash.

hobokenn: don't think so? it's in the middle of the function

YaLTeR:
Thanks!

PR #176, opened 2021-08-12
Added bxt_disable_hud / bxt_disable_vgui

SmileyAG:
Disables the HUDs in oldest games & mods (like Gunman Chronicles or Wanted!).
*`hud_draw` did not exist in earliest HLSDK versions, https://github.com/a1batross/HLSDK_ancient/blob/master/CL_DLL/HUD.CPP#L64

Big thanks to @hobokenn for helping in implementation of disabling VGUI on engineside


hobokenn:
@SmileyAG i was able to hook this function and zero it out, this is probably what you want to do it instead of byte changing, didn't test with wanted though.

SmileyAG:
> 
> 
> @SmileyAG i was able to hook this function and zero it out, this is probably what you want to do instead of byte changing, didn't test with wanted though.
> 
> image
> don't know how to disable this sprite or whatever it is, but everything else seems to be disabled
> [https://github.com/hobokenn/BunnymodXT/tree/disable-gunman-hud](https://github.com/hobokenn/BunnymodXT/tree/disable-gunman-hud?rgh-link-date=2021-08-12T18%3A27%3A06Z)

Thanks so much for implementation! Isn't not sprite, it was .tga files, you could mask out them by removing or rename `rewolf/gfx/vgui` folder

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -715,7 +734,13 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, HUD_Reset)
...
 HOOK_DEF_2(ClientDLL, void, __cdecl, HUD_Redraw, float, time, int, intermission)
 {
-	ORIG_HUD_Redraw(time, intermission);
+	switch (CVars::bxt_disable_hud.GetInt())
+	{
+		case 1:
+			break;
+		default:
+			ORIG_HUD_Redraw(time, intermission);
+	}
```
YaLTeR, line 743: that is a very roundabout way of writing `if (CVars::bxt_disable_hud.GetInt() != 1)`

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -811,3 +836,16 @@ HOOK_DEF_2(ClientDLL, int, __cdecl, HUD_UpdateClientData, client_data_t*, pcldat
...
+HOOK_DEF_4(ClientDLL, void, __fastcall, CWeaponStatusPanel__AddToHistory, void*, thisptr, int, edx, int, param1, int, param2)
+{
+	switch (CVars::bxt_disable_hud.GetInt())
+	{
+		case 1:
+			break;
+		case 2:
+			break;
+		default:
+			ORIG_CWeaponStatusPanel__AddToHistory(thisptr, edx, param1, param2);
```
YaLTeR, line 849: Hm, how does this work? Why would commenting out a function called `CWeaponStatusPanel::AddToHistory` disable HUD drawing? I guess it's related to why bxt-rs HUD upscale didn't work for it, because they draw HUD in a weird spot in the code?

YaLTeR (changes requested):
> If value == 2, its disables only HUD from GMC (useful for capturing segments)

I think it would be better to have a separate cvar for this like `bxt_disable_gmc_hud`.

BunnymodXT/modules/ClientDLL.cpp
YaLTeR: but anyway with 2 separate cvars this would be a GetBool

SmileyAG: Well, i'm kinda low-skill in coding / reverse-engineering, so feel free to correct me, i'll take a note.

And yeah, I'll just do another cvar later for this as you said.

YaLTeR (approved):

BunnymodXT/modules/ClientDLL.cpp
SmileyAG: > 
> 
> Hm, how does this work? Why would commenting out a function called `CWeaponStatusPanel::AddToHistory` disable HUD drawing? I guess it's related to why bxt-rs HUD upscale didn't work for it, because they draw HUD in a weird spot in the code?

Well, Gunman Chronicles, besides using sprites, also uses .tga files for HUDs, I don't think you hooked / modified some related function of them in bxt-rs.

Ammohistory HUD looks like really draw in a weird spot of code, cuz zero out `HUD_Redraw` and `VGUI_Label_Paint` functions doesn't rid some black semi-transparent box in left up corner of game, this is the reason why we doing zero out of history function itself to get rid that box.

hobokenn: would it be better to disable vgui on the engine side like this? https://www.youtube.com/watch?v=6rsNZ9jH_F0

YaLTeR: not sure disabling vgui is better here

hobokenn: one advantage is that it would work for other games and supersede AddToHistory and VGUI_Label_Paint

YaLTeR: so have bxt_disable_hud and bxt_disable_vgui?

hobokenn: something like that, yeah

YaLTeR (approved):

YaLTeR:
Thanks!

PR #177, opened 2021-08-12
Added bxt_hud_quickgauss

chinese-soup:
`[6:03 AM] quadkid: would it be possible to get a gpGlobals->time hud to better understand quickgauss?`


YaLTeR:
image

Should be discussed with quad as he wants the feature. One idea would be to show `max(0, 4 - gpGlobals->time)` as a countdown to when you can quickgauss.

PR #182, opened 2021-08-24
Added GL Crosshairs from OpenAG

### LINKED ISSUE #4: GL Crosshairs
Add a crosshair that is rendered by the game itself, instead of using sprite files.

Not only this would make the crosshairs look sharp and clean, but commands for full crosshair customization could also be added.

Would be best if it worked like this: http://tools.dathost.net/, so you could generate crosshairs using this, or even use your crosshair settings from CSGO and it'd look the same.

SmileyAG:
Closes #4

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -6,6 +6,7 @@
 #include "hud_custom.hpp"
 #include "interprocess.hpp"
 #include "runtime_data.hpp"
+#include "opengl_utils.hpp"
 
 #include <GL/gl.h>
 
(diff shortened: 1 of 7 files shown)


BunnymodXT/hud_custom.cpp
```diff
@@ -1116,6 +1117,145 @@ namespace CustomHud
...
+			gl.circle(center, circle_points);
+		}
+
+		// Draw the dot.
+		if (CVars::bxt_cross_dot_size.GetFloat() > 0.0f) {
+				float r = 0.0f, g = 0.0f, b = 0.0f;
+				std::istringstream ss(CVars::bxt_cross_dot_color.GetString());
+				ss >> r >> g >> b;
+
+				static float crosshairDotColor[2];
```
YaLTeR, line 1241:
```suggestion
				float crosshairDotColor[2];
```

BunnymodXT/hud_custom.cpp
```diff
@@ -1116,6 +1117,145 @@ namespace CustomHud
...
+				crosshairDotColor[0] = r;
+				crosshairDotColor[1] = g;
+				crosshairDotColor[2] = b;
+
+				if (!CVars::bxt_cross_dot_color.IsEmpty()) {
+					gl.color(crosshairDotColor[0], crosshairDotColor[1], crosshairDotColor[2], alpha);
+				} else {
+					gl.color(255.0f, 0.0f, 0.0f, alpha);
```
YaLTeR, line 1249: leave color as is if cannot parse

YaLTeR (changes requested):

BunnymodXT/hud_custom.cpp
```diff
@@ -1116,6 +1117,138 @@ namespace CustomHud
...
+				old_circle_radius = radius;
+			}
+
+			gl.circle(center, circle_points);
+		}
+
+		// Draw the dot.
+		if (CVars::bxt_cross_dot_size.GetFloat() > 0.0f) {
+			unsigned char r, g, b;
+			if (sscanf(CVars::bxt_cross_dot_color.GetString().c_str(), "%hhu %hhu %hhu", &r, &g, &b) != 3) {
```
YaLTeR, line 1237: If this fails, just don't change the color, like in OpenAG.

BunnymodXT/hud_custom.cpp
```diff
@@ -1116,6 +1117,133 @@ namespace CustomHud
 			DrawMultilineString(x, y, out.str());
 	}
 
+	static void DrawCrosshair(float time)
+	{
+		if (!CVars::bxt_cross.GetBool())
+			return;
+
+		static float old_circle_radius = 0;
+		static std::vector<Vector2D> circle_points;
```
YaLTeR, line 1126: Could you move this below to where they are used so it's easier to understand?

YaLTeR (approved):

hobokenn:
Could use `GetBool()` instead of `GetFloat()` in a few places to improve readability, for example, `if (CVars::bxt_cross_top_line.GetFloat())` and similar or `if (someCvar.GetFloat() > 0.0f)`

Issue #64, opened 2019-12-24
Prevent easy Half-Life crashes


YaLTeR:
Crashes like hornets past a level change. Catch them in BXT, and instead of crashing just stop the timer and display a message that the game would have crashed. Much nicer while not providing unfair advantage.

PR #189, opened 2021-09-03
Added hornet crash prevention

chinese-soup:
Implements part of #64 

Prints crash message to "chat" and console from as a usermsg from serverDLL so that it's visible in the demo, doesn't stop demo playback. 

**Video:**
https://user-images.githubusercontent.com/5108747/132017938-b239f6a6-e351-49df-bbca-e258688e9e80.mp4

* Tested on Windows & Linux. 
* If you think registering a usermsg could be potentially bad, then hooking `UTIL_ClientPrintAll` from ServerDLL is an alternative, if that's preferred, I'll do that instead.
* I can add a ClientDLL hud thing as well, I just didn't wanna bother (KISS) because it requires fiddling with flTime in order for it to disappear eventually, but if it'd be preferred, I'll do it.

Save file by Smiley for Steam version: https://shitpost.fun/hornet.sav (sometimes it doesn't crash when you move out of the way or the hornet just doesn't hit you because RNG, like seen in the video)
BT:
```c++
#0  CBaseEntity::Instance (pev=0x0) at ../dlls/player.cpp:4901
#1  CBasePlayer::TakeDamage (this=0x9d815f0, pevInflictor=0xc862df04, pevAttacker=0x0, flDamage=4, bitsDamageType=2) at ../dlls/player.cpp:471
#2  0x8630daef in CHornet::DieTouch (this=0x9d61bc0, pOther=0x9d815f0) at ../dlls/hornet.cpp:410
#3  0xf4b35075 in SV_Impact (ptrace=0xffec0b60, e2=0xc861076c, e1=0xc862de84) at ../engine/sv_phys.c:195
#4  SV_Impact (e1=0xc862de84, e2=0xc861076c, ptrace=0xffec0b60) at ../engine/sv_phys.c:170
#5  0xf4b35ac2 in SV_PushEntity (ent=0xc862de84, push=0xffec0c80) at ../engine/sv_phys.c:545
#6  0xf4b3803a in SV_Physics_Toss (ent=0xc862de84) at ../engine/sv_phys.c:1578
#7  0xf4b396e8 in SV_Physics () at ../engine/sv_phys.c:2054
#8  0xf4b31236 in SV_Frame () at ../engine/sv_main.c:9355
#9  0xf4af07c6 in _Host_Frame (time=0.0216090735) at ../engine/host.c:1430
#10 0xf4af0c52 in Host_Frame (time=0.0216090735, iState=4, stateInfo=0xffec0e0c) at ../engine/host.c:1548
#11 0xf4b1db04 in CEngine::Frame (this=0xf4d37aa0 <g_Engine>) at ../engine/sys_engine.cpp:245
```

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -64,6 +64,11 @@ extern "C" void __cdecl _ZN12CTriggerSave9SaveTouchEP11CBaseEntity(void* thisptr
 {
 	return ServerDLL::HOOKED_CTriggerSave__SaveTouch_Linux(thisptr, pOther);
 }
+
+extern "C" int __cdecl _ZN11CBasePlayer10TakeDamageEP9entvars_sS1_fi(void* thisptr, entvars_t* pevInflictor, entvars_t* pevAttacker, float flDamage, int bitsDamageType)
+{
+	return ServerDLL::HOOKED_CBasePlayer__TakeDamage_Linux(thisptr, pevInflictor, pevAttacker, flDamage, bitsDamageType);
+}
 #endif
 
 void ServerDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* moduleBase, size_t moduleLength, bool needToIntercept)
@@ -181,6 +186,7 @@ void ServerDLL::Clear()
 	ORIG_ClientCommand = nullptr;
 	ORIG_CPushable__Move = nullptr;
 	ORIG_CBasePlayer__TakeDamage = nullptr;
+	ORIG_CBasePlayer__TakeDamage_Linux = nullptr;
 	ORIG_GetEntityAPI = nullptr;
 	ORIG_CGraph__InitGraph = nullptr;
 	ORIG_CGraph__InitGraph_Linux = nullptr;
@@ -665,8 +671,13 @@ void ServerDLL::FindStuff()
 		if (ORIG_CBasePlayer__TakeDamage) {
 			EngineDevMsg("[server dll] Found CBasePlayer::TakeDamage at %p (using the %s pattern).\n", ORIG_CBasePlayer__TakeDamage, pattern->name());
 		} else {
-			EngineDevWarning("[server dll] Could not find CBasePlayer::TakeDamage.\n");
-			EngineWarning("Damage logging is not available.\n");
+			ORIG_CBasePlayer__TakeDamage_Linux = reinterpret_cast<_CBasePlayer__TakeDamage_Linux>(MemUtils::GetSymbolAddress(m_Handle, "_ZN11CBasePlayer10TakeDamageEP9entvars_sS1_fi"));
+			if (ORIG_CBasePlayer__TakeDamage_Linux) {
+				EngineDevMsg("[server dll] Found CBasePlayer::TakeDamage_Linux at %p.\n", ORIG_CBasePlayer__TakeDamage_Linux);
+			} else {
+				EngineDevWarning("[server dll] Could not find CBasePlayer::TakeDamage.\n");
+				EngineWarning("Damage logging is not available.\n");
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1825,6 +1836,23 @@ HOOK_DEF_4(ServerDLL, void, __fastcall, CPushable__Move, void*, thisptr, int, ed
 	ORIG_CPushable__Move(thisptr, edx, pOther, push);
 }
 
+void ServerDLL::DoWouldCrashMessage()
+{
+	// We send a user message from the ServerDLL so that we have the notice in the demo.
+	int gmsgTextMsg = pEngfuncs->pfnRegUserMsg( "TextMsg", -1 );
+
+	pEngfuncs->pfnMessageBegin(MSG_ALL, gmsgTextMsg, NULL, NULL);
+	pEngfuncs->pfnWriteByte(HUD_PRINTTALK );
```
YaLTeR, line 1845: how'd you manage this one )

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1825,6 +1836,23 @@ HOOK_DEF_4(ServerDLL, void, __fastcall, CPushable__Move, void*, thisptr, int, ed
 	ORIG_CPushable__Move(thisptr, edx, pOther, push);
 }
 
+void ServerDLL::DoWouldCrashMessage()
+{
+	// We send a user message from the ServerDLL so that we have the notice in the demo.
+	int gmsgTextMsg = pEngfuncs->pfnRegUserMsg( "TextMsg", -1 );
```
YaLTeR, line 1842: I'm assuming this is one of the standard usermsgs, does registering it return the same int or something? Is it ok to register every time?

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1855,7 +1883,29 @@ HOOK_DEF_6(ServerDLL, int, __fastcall, CBasePlayer__TakeDamage, void*, thisptr,
 		HwDLL::GetInstance().logWriter.PushDamage(damage);
 	}
 
-	return ORIG_CBasePlayer__TakeDamage(thisptr, edx, pevInflictor, pevAttacker, flDamage, bitsDamageType);
+	if (pevAttacker == nullptr)
```
YaLTeR, line 1886: Is it more or less a given that the game will crash with null attacker? Obviously mods can add whatever code but if there's nothing complicated in this function then it should be ok I suppose

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1855,7 +1883,29 @@ HOOK_DEF_6(ServerDLL, int, __fastcall, CBasePlayer__TakeDamage, void*, thisptr,
...
+	{
+		DoWouldCrashMessage();
+		return 0;
+	}
+	else
+	{
+		return ORIG_CBasePlayer__TakeDamage_Linux(thisptr, pevInflictor, pevAttacker, flDamage, bitsDamageType);
+	}
+
```
YaLTeR, line 1908: wouldn't be a chinese-soup PR

YaLTeR (commented):
> If you think registering a usermsg could be potentially bad, then hooking UTIL_ClientPrintAll from ServerDLL is an alternative, if that's preferred, I'll do that instead.

I guess the only bad thing is if it's handled in the client and then some mod removes or overrides it but I'm not sure this is as big of a concern with a usermsg as generic as showing text in chat

BunnymodXT/modules/ServerDLL.cpp
chinese-soup: Was originally Copy paste and then I fixed it up but apparently didn't finish 😂😂

BunnymodXT/modules/ServerDLL.cpp
hobokenn: i know of at least one other crash like that https://github.com/ValveSoftware/halflife/issues/3056

https://user-images.githubusercontent.com/50155487/132092502-c7c8cb1a-99c9-493c-9839-46119d5cd429.mp4

chinese-soup: @YaLTeR As you've said, mods can add anything they want, should I therefore add a check for "op4" and "hl" since those are the ones we are sure will always crash on TakeDamage, because of CBaseEntity::Instance?

```c++
// dlls/player.cpp
int CBasePlayer :: TakeDamage( entvars_t *pevInflictor, entvars_t *pevAttacker, float flDamage, int bitsDamageType )
{
	// [...] omitted
	
	// go take the damage first

	CBaseEntity *pAttacker = CBaseEntity::Instance(pevAttacker);```

The specific overloaded function for CBaseEntity::Instance called from TakeDamage is the `entvars_t` one:
```c++
// cbase.h
static CBaseEntity *Instance( entvars_t *pev ) { return Instance( ENT( pev ) ); }
```
`ENT` is defined as 
```c++
// util.h
inline edict_t *ENT(const entvars_t *pev)	{ return pev->pContainingEntity; }
```
Which causes the crash becaues the pev pointer is null.

I don't really know if CBaseEntity::Instance() is actually hookable, I am not C++ proficient enough to know, it's in cbase.h, if anyone smarter wants to check (this is also why I hooked TakeDamage instead).

BunnymodXT/modules/ServerDLL.cpp
chinese-soup: Replaced with a simple server cmd "say" so that it's portable and just silently fails if "say" isn't a command in some mod.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1838,14 +1838,9 @@ HOOK_DEF_4(ServerDLL, void, __fastcall, CPushable__Move, void*, thisptr, int, ed
...
 {
-	// We send a user message from the ServerDLL so that we have the notice in the demo.
-	int gmsgTextMsg = pEngfuncs->pfnRegUserMsg( "TextMsg", -1 );
-
-	pEngfuncs->pfnMessageBegin(MSG_ALL, gmsgTextMsg, NULL, NULL);
-	pEngfuncs->pfnWriteByte(HUD_PRINTTALK );
-	pEngfuncs->pfnWriteString("The game would have crashed. BXT prevented the crash and the timer was stopped. This run is no longer valid.\n");
-	pEngfuncs->pfnMessageEnd();
-
+	char command[] = "say The game would have crashed. BXT prevented the crash and the timer was stopped. This run is no longer valid. \n";
```
YaLTeR, line 1841: Why the trailing space?

BunnymodXT/modules/ServerDLL.cpp
YaLTeR: Eh let's leave it as is, at least if it triggers somewhere unexpected we get a loud message in the chat.

BunnymodXT/modules/ServerDLL.cpp
chinese-soup: Fixed

YaLTeR:
Thanks!

PR #193, opened 2021-09-08
Added op4won pattern for CBaseMonster__Killed

SmileyAG:
In case, if some older Opposing Force based modification ends on death of NPC, it will be cool to have ready-found function for implement autostop.


YaLTeR:
Thanks

PR #194, opened 2021-09-15
Added bxt_water_remove

SmileyAG:
Boosts framerate via disabling draw of water textures*

*Yes I know that water could be disabled via `r_drawentities 0`, although its not works with all types of water (e.g., `map c1a4`)

I decided to not add check for cheats on the cvar, since I use this for segmenting and I would not like to see mention of `sv_cheats 1` in the log of .dem 

UPD: Recommended to use in combination with `r_novis 1` to make water fully masked

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -198,6 +198,11 @@ extern "C" qboolean __cdecl BIsValveGame()
 {
 	return true;
 }
+
+extern "C" void __cdecl EmitWaterPolys(msurface_t *fa, int direction)
+{
+	return HwDLL::HOOKED_EmitWaterPolys(fa, direction);
+}
 #endif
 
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4732,7 +4752,7 @@ HOOK_DEF_0(HwDLL, void, __cdecl, R_Clear)
 {
 	// This is needed or everything will look washed out or with unintended
 	// motion blur.
-	if (CVars::sv_cheats.GetBool() && (CVars::bxt_wallhack.GetBool() || CVars::bxt_skybox_remove.GetBool())) {
+	if (CVars::bxt_water_remove.GetBool() || CVars::sv_cheats.GetBool() && (CVars::bxt_wallhack.GetBool() || CVars::bxt_skybox_remove.GetBool())) {
```
YaLTeR, line 4755: Please use brackets here because the order of `||` and `&&` operators is always unclear.

YaLTeR (approved):

YaLTeR:
Thanks

PR #203, opened 2021-10-03
Hud extensions

Eddio0141:
Quite a while back I crashed my game and it was to do in triangle_drawing.cpp around line 663 but I couldn't recreate the crash and couldn't think of a fix. I might have accidentally fixed it but I'm not sure.

Health and armor data is now being sent from the client and is working, but tas editor status shows you have 0 hp and armor by default until simulation updates it. I'm not sure if that's a good way to say the server doesn't know what health and armor it has at whatever framebulk selected.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -767,7 +767,7 @@ namespace CustomHud
 		if (CVars::bxt_hud_selfgauss.GetBool())
 		{
 			int x, y;
-			GetPosition(CVars::bxt_hud_selfgauss_offset, CVars::bxt_hud_selfgauss_anchor, &x, &y, -200, (si.iCharHeight * 30) + 3);
+			GetPosition(CVars::bxt_hud_selfgauss_offset, CVars::bxt_hud_selfgauss_anchor, &x, &y, -200, (si.iCharHeight * 26) + 3);
 
 			bool selfgaussable;
 			int hitGroup = 0; // It's always initialized if selfgaussable is set to true, but GCC issues a warning anyway.
@@ -827,7 +827,7 @@ namespace CustomHud
 		if (CVars::bxt_hud_nihilanth.GetBool())
 		{
 			int x, y;
-			GetPosition(CVars::bxt_hud_nihilanth_offset, CVars::bxt_hud_nihilanth_anchor, &x, &y, -200, (si.iCharHeight * 33) + 3);
+			GetPosition(CVars::bxt_hud_nihilanth_offset, CVars::bxt_hud_nihilanth_anchor, &x, &y, -200, (si.iCharHeight * 29) + 3);
 
 			std::ostringstream out;
 			out << "Nihilanth:\n";
@@ -861,7 +861,7 @@ namespace CustomHud
 		if (CVars::bxt_hud_gonarch.GetBool())
 		{
 			int x, y;
-			GetPosition(CVars::bxt_hud_gonarch_offset, CVars::bxt_hud_gonarch_anchor, &x, &y, -200, (si.iCharHeight * 40) + 3);
+			GetPosition(CVars::bxt_hud_gonarch_offset, CVars::bxt_hud_gonarch_anchor, &x, &y, -200, (si.iCharHeight * 36) + 3);
 
 			std::ostringstream out;
 			out << "Gonarch:\n";
@@ -1034,7 +1034,7 @@ namespace CustomHud
 			return;
 
 		int x, y;
-		GetPosition(CVars::bxt_hud_tas_editor_status_offset, CVars::bxt_hud_tas_editor_status_anchor, &x, &y, -250, (si.iCharHeight * 43) + 3);
+		GetPosition(CVars::bxt_hud_tas_editor_status_offset, CVars::bxt_hud_tas_editor_status_anchor, &x, &y, -250, (si.iCharHeight * 39) + 3);
 
 		std::ostringstream out;
 		out.setf(std::ios::fixed);
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -696,13 +696,13 @@ namespace CustomHud
 					out << targetname << '\n';
 				}
 
-				out << "HP: " << ent->v.health << "\n\n";
+				out << "HP: " << ent->v.health << '\n';
 
-				out << "Yaw: " << ent->v.angles[1] << "\n\n";
+				out << "Yaw: " << ent->v.angles[1] << '\n';
 
 				out << "X: " << ent->v.origin.x << '\n';
 				out << "Y: " << ent->v.origin.y << '\n';
-				out << "Z: " << ent->v.origin.z << "\n\n";
+				out << "Z: " << ent->v.origin.z << '\n';
 
 				out << "X Vel: " << ent->v.velocity.x << '\n';
 				out << "Y Vel: " << ent->v.velocity.y << '\n';
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -71,15 +71,16 @@ namespace CustomHud
 		bool yaw_present;
 
 		std::string commands;
+
+		float vel;
+		float zvel;
+		float zpos;
+		float realyaw;
+		float health;
+		float armor;
 	};
 	static FrameBulkStatus frame_bulk_status;
-	static float player_vel_status;
-	static float player_zvel_status;
-	static float player_zpos_status;
-	static float player_realyaw_status;
 	static bool frame_bulk_selected;
-	static float player_health_status;
-	static float player_armor_status;
 
 	template<typename T, size_t size = 3>
 	static inline void vecCopy(const T src[], T dest[])
@@ -1110,18 +1111,18 @@ namespace CustomHud
 
 			out << '\n';
 
-			out << "Camera Yaw: " << player_realyaw_status << '\n';
+			out << "Camera Yaw: " << frame_bulk_status.realyaw << '\n';
 
 			out << '\n';
 
-			out << "Vel: " << player_vel_status << '\n';
-			out << "Z Vel: " << player_zvel_status << '\n';
-			out << "Z Pos: " << player_zpos_status << '\n';
+			out << "Vel: " << frame_bulk_status.vel << '\n';
+			out << "Z Vel: " << frame_bulk_status.zvel << '\n';
+			out << "Z Pos: " << frame_bulk_status.zpos << '\n';
 
 			out << '\n';
 
-			out << "Health: " << player_health_status << '\n';
-			out << "Armor: " << player_armor_status << '\n';
+			out << "Health: " << frame_bulk_status.health << '\n';
+			out << "Armor: " << frame_bulk_status.armor << '\n';
 		} else {
 			out << " no frame bulk selected";
 		}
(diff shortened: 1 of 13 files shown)


SmileyAG:
You didn't changed `ry` offsets of existed HUDs and now it overlay on them:

изображение

Enable all HUDs: `bxt_hud_speedometer 0;bxt_hud_distance 1;bxt_hud_entity_info 1;bxt_hud_gonarch 1;bxt_hud_nihilanth 1;bxt_hud_origin 1;bxt_hud_selfgauss 1;bxt_hud_tas_editor_status 1;bxt_hud_velocity 1;bxt_hud_viewangles 1;bxt_hud_visible_landmarks 1`

YaLTeR:
Maybe we should move default TAS editor status position somewhere else so there's no overlap problem with the right side HUD stack?

BunnymodXT/hud_custom.cpp
```diff
@@ -1066,6 +1107,21 @@ namespace CustomHud
 			if (!frame_bulk_status.commands.empty()) {
 				out << "Commands:\n  " << frame_bulk_status.commands << '\n';
 			}
+
+			out << '\n';
+
+			out << "Real Yaw: " << player_realyaw_status << '\n';
```
YaLTeR, line 1113: "Yaw" vs. "Real Yaw" is confusing. I suggest something like "Strafing Yaw" and "Camera Yaw"

BunnymodXT/simulation_ipc.cpp
```diff
@@ -40,7 +40,7 @@ namespace simulation_ipc {
 	};
 
 	enum ClientToServerMessageType {
-		SIMULATED_FRAME = 0,
+		SIMULATED_FRAME = 0
```
YaLTeR, line 43: Trailing commas are there for a reason—you can add or remove lines without changing the previous line, which clutters the diff and leads to conflicts. Please leave them.

BunnymodXT/triangle_drawing.cpp
```diff
@@ -651,8 +651,39 @@ namespace TriangleDrawing
 		if (closest_edge_frame != 0) {
 			auto& frame_bulk = input.frame_bulks[closest_edge_prev_frame_bulk_index];
 
-			// Update the HUD status before any changes, since that's the state that was visualized earlier.
-			CustomHud::UpdateTASEditorStatus(frame_bulk);
+				uint64_t current_frame = 0;
+				for (size_t i = 0; i < closest_edge_prev_frame_bulk_index + 1; i++) {
+					current_frame += input.frame_bulks[i].GetRepeats();
```
YaLTeR, line 656: I believe closest frame is already computed above for use in insert_point.

BunnymodXT/triangle_drawing.cpp
```diff
@@ -651,8 +651,39 @@ namespace TriangleDrawing
...
+				}
+
+				auto data_size = input.player_datas.size();
+
+				// TODO: fix some rare? crash that might still exist
```
YaLTeR, line 661: The crash is probably that when the part has not yet been predicted, player_datas may not have current_frame.

BunnymodXT/input_editor.cpp
```diff
@@ -210,6 +210,12 @@ void EditedInput::mark_as_stale(size_t frame_bulk_index) {
 	frame_bulk_starts.erase(frame_bulk_starts.begin() + frame_bulk_index + 1, frame_bulk_starts.end());
 
 	auto first_frame = std::min(*(frame_bulk_starts.cend() - 1), player_datas.size() - 1);
+	if (player_health_datas.size() > 1) {
```
YaLTeR, line 213: Why are these special-cased like this? I'd prefer that they are handled in exactly the same way as all other vectors here.

BunnymodXT/hud_custom.cpp
```diff
@@ -73,7 +73,13 @@ namespace CustomHud
 		std::string commands;
 	};
 	static FrameBulkStatus frame_bulk_status;
+	static float player_vel_status;
+	static float player_zvel_status;
+	static float player_zpos_status;
+	static float player_realyaw_status;
 	static bool frame_bulk_selected;
+	static float player_health_status;
+	static float player_armor_status;
```
YaLTeR, line 82: Just add these into FrameBulkStatus and maybe rename it to something like TASEditorStatus

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3220,6 +3228,8 @@ void HwDLL::InsertCommands()
 					PrevFraction,
 					PrevNormalz,
 					thisFrameIs0ms,
+					health,
+					armor
```
YaLTeR, line 3232: Leave trailing comma

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2023,6 +2023,23 @@ bool ServerDLL::GetNihilanthInfo(float &health, int &level, int &irritation, boo
 	return true;
 }
 
+bool ServerDLL::GetGonarchInfo(float &health, int& sequence, float& frame) const {
+	// TODO: figure out what infos are needed
+	// needed: m_nodeTime
```
YaLTeR, line 2028: Do you want to resolve this before merging? Otherwise just remove the comment, we can add the info as we need it.

BunnymodXT/triangle_drawing.cpp
```diff
@@ -651,8 +651,39 @@ namespace TriangleDrawing
 		if (closest_edge_frame != 0) {
 			auto& frame_bulk = input.frame_bulks[closest_edge_prev_frame_bulk_index];
 
-			// Update the HUD status before any changes, since that's the state that was visualized earlier.
-			CustomHud::UpdateTASEditorStatus(frame_bulk);
+				uint64_t current_frame = 0;
```
YaLTeR, line 654: The indentation is 1 tab more than needed.

YaLTeR (changes requested):

Eddio0141:
ok this should be everything now

BunnymodXT/input_editor.cpp
```diff
@@ -347,8 +343,21 @@ void EditedInput::received_simulated_frame(const simulation_ipc::SimulatedFrame
 	if (player_armor_datas.size() == 0) {
 		player_armor_datas.push_back(0);
 	}
-	player_health_datas.push_back(frame.health);
-	player_armor_datas.push_back(frame.armor);
+
+	// only push back if the simulated frame doesn't exist
```
YaLTeR, line 347: Just push zeros in simulate() along with others, so all vectors are handled in exactly the same way. This significantly reduces potential for mistakes.

BunnymodXT/simulation_ipc.cpp
```diff
@@ -280,7 +280,7 @@ namespace simulation_ipc {
 
 		ClientToServerMessage message {
 			ClientToServerMessageType::SIMULATED_FRAME,
-			frame,
```
YaLTeR, line 283: give comma back pls

BunnymodXT/triangle_drawing.cpp
```diff
@@ -651,8 +651,39 @@ namespace TriangleDrawing
...
 			auto& frame_bulk = input.frame_bulks[closest_edge_prev_frame_bulk_index];
 
+			auto data_size = input.player_datas.size();
+
+			float current_player_vel = 0;
+			float current_player_zvel = 0;
+			float current_player_zpos = 0;
+			float current_player_realyaw = 0;
+
+			if (data_size >= closest_edge_frame) {
```
YaLTeR, line 661: This should be >. `vector[vector.size()]` is accessing one value past the end.

YaLTeR: Also why is data_size defined above instead of right in the condition?

BunnymodXT/triangle_drawing.cpp
```diff
@@ -651,8 +651,39 @@ namespace TriangleDrawing
...
+				current_player_vel = std::hypotf(current_player_vels[0], current_player_vels[1]);
+				current_player_zvel = current_player_vels[2];
+				current_player_zpos = current_player_data.Origin[2];
+				current_player_realyaw = current_player_data.Viewangles[1];
+			}
+
+			float current_player_health = 0;
+			float current_player_armor = 0;
+
+			if (input.player_health_datas.size() > closest_edge_frame) {
```
YaLTeR, line 674: You won't need the separate condition once you unify the vector handling.

BunnymodXT/input_editor.cpp
```diff
@@ -253,6 +261,10 @@ void EditedInput::set_repeats(size_t frame_bulk_index, unsigned repeats) {
 	if (last_frame >= player_datas.size() - 2)
 		return;
 
+	if (player_health_datas.size() > last_frame + 1) {
```
YaLTeR, line 264: This check is no longer needed because they are handled the same way as others.

BunnymodXT/triangle_drawing.cpp
YaLTeR: Yeah, now it's no longer needed, this can go into the condition at the top.

YaLTeR (approved):

chinese-soup:
Hello, this still cuts into bxt_hud_selfgauss by default when pointing at an entity that has a name.

Entity with targetname:
image
Entity with no targetname (e.g. worldspawn)
image

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2023,6 +2023,20 @@ bool ServerDLL::GetNihilanthInfo(float &health, int &level, int &irritation, boo
 	return true;
 }
 
+bool ServerDLL::GetGonarchInfo(float &health, int& sequence, float& frame) const {
```
chinese-soup, line 2026:
```suggestion
bool ServerDLL::GetGonarchInfo(float &health, int &sequence, float& frame, std::string& sequence_string) const {
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2023,6 +2023,20 @@ bool ServerDLL::GetNihilanthInfo(float &health, int &level, int &irritation, boo
...
+	// There should be only 1 big momma
+	edict_t* pent = pEngfuncs->pfnFindEntityByString(nullptr, "classname", "monster_bigmomma");
+	if (!pent || !pEngfuncs->pfnEntOffsetOfPEntity(pent)) {
+		return false;
+	}
+
+	health = pent->v.health;
+	sequence = pent->v.sequence;
+	frame = pent->v.frame;
+
```
chinese-soup, line 2036:
```suggestion
sequence_string = std::string(GetString(pent->v.netname));

```

BunnymodXT/modules/ServerDLL.hpp
```diff
@@ -62,6 +62,8 @@ class ServerDLL : public IHookableDirFilter
 	std::vector<const Vector *> GetDisplacerTargets() const;
 	bool GetNihilanthInfo(float &health, int &level, int &irritation, bool &recharger, int &nspheres, int &sequence, float &frame) const;
 	
+	bool GetGonarchInfo(float& health, int& sequence, float& frame) const;
```
chinese-soup, line 65:
```suggestion
	bool GetGonarchInfo(float& health, int& sequence, float& frame, std::string& sequence_string) const;
```

chinese-soup (changes requested):
How about something like this for the stringified sequence? (If this is what you meant, from `dlls/bigmomma.cpp:int GetNodeSequence( void )`)

Eddio0141:
uhh how can i pull the latest commit from here so I can test it and fix something going wrong?

YaLTeR:
Thanks!

PR #204, opened 2021-10-12
Add cvars to disable viewmodel idle/equip sequences

hobokenn:
Based on the OpenAG implementation, thanks @chinese-soup.
Engine hook for old WON mods, everything else uses client's function


hobokenn:
FWIW i was able to call a function which resets a pointer to studioapi to point to the engine's studio functions, not sure if it introduces any bugs or breaks any mods https://github.com/hobokenn/BunnymodXT/commit/d4b1fc856fbe7ece7edada7dedfaf3544c9ee77f

PR #213, opened 2021-10-29
Added bxt_force_duck

chinese-soup:
For +duck category without them pausing cheats. https://www.speedrun.com/hl1ce#%2Bduck

hello @hobokenn, my patterns are probably all wrong and stupid, I couldn't test the WON one at all (no win machine, VM hates me with sw.dll bs error), the Steam one only matched for ClientDLL for some reason (even though in Ghidra it finds it in both Client and Server DLLs), I commented/removed the pattern code out in client DLL because for some reason having FindAsync() crashed the game on Linux. :| Please end my misery.

> [17:22] although, i can't get my Steampipe pattern to work on ServerDLL (only ClientDLL) and I can't get my WON to work at all, so I can't test it there at all, ummm calling 800-hoboken :CoolStoryBob:
> [17:22] not using ppmove in some other more top-level PM_xxx function becauase that would require pm_shared/ engine headers and those are in conflict with the linux hooks etc. etc. so fuck that, easiest & cleanest solution is imho this one, aka just hooking PM_UnDuck and not calling it at all if +duck meme run cvar is on, it's a meme category anyway
> [17:23] tiem to check for spaces b4 PR LMAO
> [17:24] also the ClientDLL hook is pretty much useless imo, for this +duck usecase, anyway
> [17:24] cuz when you pause/unpause only the ServerDLL's PM_UnDuck is called

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -803,6 +803,7 @@ void ServerDLL::FindStuff()
 	ORIG_AddToFullPack = reinterpret_cast<_AddToFullPack>(MemUtils::GetSymbolAddress(m_Handle, "_Z13AddToFullPackP14entity_state_siP7edict_sS2_iiPh"));
 	ORIG_ClientCommand = reinterpret_cast<_ClientCommand>(MemUtils::GetSymbolAddress(m_Handle, "_Z13ClientCommandP7edict_s"));
 	ORIG_PM_Move = reinterpret_cast<_PM_Move>(MemUtils::GetSymbolAddress(m_Handle, "PM_Move"));
+
 	if (ORIG_CmdStart && ORIG_AddToFullPack && ORIG_ClientCommand && ORIG_PM_Move) {
 		EngineDevMsg("[server dll] Found CmdStart at %p.\n", ORIG_CmdStart);
 		EngineDevMsg("[server dll] Found AddToFullPack at %p.\n", ORIG_AddToFullPack);
@@ -1149,6 +1150,8 @@ void ServerDLL::RegisterCVarsAndCommands()
 		REG(bxt_disable_autosave);
 	if (ORIG_CChangeLevel__UseChangeLevel && ORIG_CChangeLevel__TouchChangeLevel)
 		REG(bxt_disable_changelevel);
+	if (ORIG_PM_PlayerMove)
+		REG(bxt_force_duck);
 	#undef REG
 }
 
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -803,6 +807,8 @@ void ServerDLL::FindStuff()
 	ORIG_AddToFullPack = reinterpret_cast<_AddToFullPack>(MemUtils::GetSymbolAddress(m_Handle, "_Z13AddToFullPackP14entity_state_siP7edict_sS2_iiPh"));
 	ORIG_ClientCommand = reinterpret_cast<_ClientCommand>(MemUtils::GetSymbolAddress(m_Handle, "_Z13ClientCommandP7edict_s"));
 	ORIG_PM_Move = reinterpret_cast<_PM_Move>(MemUtils::GetSymbolAddress(m_Handle, "PM_Move"));
+	ORIG_PM_UnDuck = reinterpret_cast<_PM_UnDuck>(MemUtils::GetSymbolAddress(m_Handle, "PM_UnDuck"));
+
```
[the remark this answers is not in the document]
hobokenn: this line is unnecessary anyway, `FindFunctionAsync` already calls `GetSymbolAddress` internally

YaLTeR:
I was thinking of more something like adding the duck button bit in `CL_CreateMove` unconditionally. Then there wouldn't be any problems with `PM_UnDuck` client/server hooking and whatnot and also with any custom mod logic. Would that work?

hobokenn:
> 
> 
> I was thinking of more something like adding the duck button bit in `CL_CreateMove` unconditionally. Then there wouldn't be any problems with `PM_UnDuck` client/server hooking and whatnot and also with any custom mod logic. Would that work?

if you meant something like this, it doesn't work (crouches, but uncrouches on unpause)
```
HOOK_DEF_3(ClientDLL, void, __cdecl, CL_CreateMove, float, frametime, usercmd_s*, cmd, int, active)
{
	ORIG_CL_CreateMove(frametime, cmd, active);

	if (CVars::bxt_force_duck.GetBool())
		cmd->buttons |= IN_DUCK;
}
```

YaLTeR:
> > I was thinking of more something like adding the duck button bit in `CL_CreateMove` unconditionally. Then there wouldn't be any problems with `PM_UnDuck` client/server hooking and whatnot and also with any custom mod logic. Would that work?
> 
> if you meant something like this, it doesn't work (crouches, but uncrouches on unpause)
> 
> ```
> HOOK_DEF_3(ClientDLL, void, __cdecl, CL_CreateMove, float, frametime, usercmd_s*, cmd, int, active)
> {
> 	ORIG_CL_CreateMove(frametime, cmd, active);
> 
> 	if (CVars::bxt_force_duck.GetBool())
> 		cmd->buttons |= IN_DUCK;
> }
> ```

Hm, I wonder what makes it do that then.

chinese-soup:
Yeah, what hoboken said. Although I personally didn't even test it, because I saw that DUCK is never actually unpressed on pause/unpause when I debugged it, so I didn't even try to force the duck bit.

hobokenn:
toggling the duck bit in `PM_PlayerMove` hook before the original function runs works for me.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1305,6 +1308,9 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, PM_PlayerMove, qboolean, server)
 		LogPlayerMove(true, pmove);
 	}
 
+	if (CVars::bxt_force_duck.GetBool())
+		cmd->buttons |= IN_DUCK;
+
```
YaLTeR, line 1313: I think it would make sense to move this before the logging.

chinese-soup: aight

YaLTeR:
Thanks

PR #215, opened 2021-11-03
Added bxt_show_hidden_entities_clientside

### LINKED ISSUE #143: bxt_show_hidden_entities get saved in demos

chinese-soup:
Fixes #143

https://user-images.githubusercontent.com/5108747/140078906-ab4c2a6f-0728-4cdd-8573-5d7e3e88c185.mp4

* no time, no energy, no windows = no windows, PRs welcome (otherwise imma do it later myself)
* lemme know if renderamt > 0 check like in serverDLL version (where bxt_show_hidden_entities 1 means renderamt > 0 entities aren't set to kRenderNormal AND bxt_show_hidden_entities 2 means even renderamt > 0 are set to kRenderNormal) is preferred here as well
* too lazy to implement the legacy triggers part as it is in ServerDLL, because ent in HUD_AddEntity has no classname, meh

### DIFF
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -50,6 +50,7 @@ namespace CVars
 	CVarWrapper bxt_bhopcap("bxt_bhopcap", "1");
 	CVarWrapper bxt_timer_autostop("bxt_timer_autostop", "1");
 	CVarWrapper bxt_show_hidden_entities("bxt_show_hidden_entities", "0");
+	CVarWrapper bxt_show_hidden_entities_clientside("bxt_show_hidden_entities_clientside", "0");
 	CVarWrapper bxt_show_triggers_legacy("bxt_show_triggers_legacy", "0");
 	CVarWrapper bxt_show_pickup_bbox("bxt_show_pickup_bbox", "0");
 	CVarWrapper bxt_disable_autosave("bxt_disable_autosave", "0");
@@ -187,6 +188,7 @@ namespace CVars
 		&bxt_bhopcap,
 		&bxt_timer_autostop,
 		&bxt_show_hidden_entities,
+		&bxt_show_hidden_entities_clientside,
 		&bxt_autojump_prediction,
 		&bxt_bhopcap_prediction,
 		&bxt_show_nodes,
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -163,6 +163,7 @@ namespace CVars
 	extern CVarWrapper bxt_bhopcap;
 	extern CVarWrapper bxt_timer_autostop;
 	extern CVarWrapper bxt_show_hidden_entities;
+	extern CVarWrapper bxt_show_hidden_entities_clientside;
 	extern CVarWrapper bxt_show_triggers_legacy;
 	extern CVarWrapper bxt_show_pickup_bbox;
 	extern CVarWrapper bxt_disable_autosave;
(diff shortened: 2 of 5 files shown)


chinese-soup:
never mind, i gotta combien it with the serverDLL like i had before i deleted it all 5 minutes ago, because this only works for renderamt=0 entities

hobokenn:
@chinese-soup https://github.com/chinese-soup/BunnymodXT/pull/5

SmileyAG:
@chinese-soup What about rename ServerDLL implementation with `bxt_show_hidden_entities_legacy` cmd, as it done for triggers?

chinese-soup:
@SmileyAG 
> @chinese-soup What about rename ServerDLL implementation with `bxt_show_hidden_entities_legacy` cmd, as it done for triggers?
> 
> Because I don't think that someone would be still to use ServerDLL implementation, since it's being visible in demos....

That's not a good idea.

See my notes above, the thing is, clientside hidden entites that can be shown WITHOUT changing what the server sends are only the ones who have renderamt set to 0 (aka they are "fully transparent"), so I can make this Cvar do only this, because if I were to draw EF_NODRAW clientside then the thing is that the entities would still be visible in the demo if you played back the demo without BXT (since I'd need to unset the EF_NODRAW flag from those and send them to the ClientDLL and "mark" them soemhow (e.g. renderfx = 250) and then substract the 250 in ClientDLL hook, but then obviously the (agian, for example) renderfx would be changed in the demo itself and to get the proper renderfx you'd need to watch the demo with BXT (where the Client would change it again)... 

Hope this is udnerstandable somewhat, lol.

Basically: if anyone can show me some EF_NODRAW entities examples in-game, that'd be nice and I will try to figure something out, otherwise I might just give up and make this only work for renderamt/rendermode hidden entities since I don't feel like modifying the serverDLL entities in order to change them back in the clientDLL hook based on a cvar isn't really a good idea. --> Therefore the ServerDLL cvar would still be useful, because of EF_NODRAW entities.

SmileyAG:
> 
> 
> @SmileyAG
> 
> > @chinese-soup What about rename ServerDLL implementation with `bxt_show_hidden_entities_legacy` cmd, as it done for triggers?
> > Because I don't think that someone would be still to use ServerDLL implementation, since it's being visible in demos....
> 
> That's not a good idea.
> 
> See my notes above, the thing is:
> 
> Clientside hidden entites that can be shown WITHOUT changing what the server sends are only the ones who have renderamt set to 0 (aka they are "fully transparent"), so I can make this cvar do only this.
> 
> Because if I were to draw EF_NODRAW clientside then the thing is that the entities would still be visible in the demo if you played back the demo without BXT (since I'd need to unset the EF_NODRAW flag from those in ServerDLL and send them to the ClientDLL and "mark" them as "changed" (i.e. changing them in the demo as well) somehow (e.g. renderfx = 250) and then substract the 250 in ClientDLL hook, but then obviously the (agian, for example) renderfx would be changed in the demo itself and to get the proper renderfx you'd need to watch the demo with BXT (where the Client would change it during demo playback)...
> 
> Hope this is understandable somewhat, lol.
> 
> Basically: if anyone can show me some EF_NODRAW entities examples in-game, that'd be nice and I will try to figure something out, otherwise I might just give up and make this only work for renderamt/rendermode hidden entities since I don't feel like modifying the serverDLL entities in order to change them back in the clientDLL hook based on a cvar is a good idea. --> Therefore the ServerDLL cvar would still be useful, because of EF_NODRAW entities.

Got it. 

Alright, after a while I will try to give the examples of maps where entities have an `EF_NODRAW` flag.

YaLTeR:
Maybe something like change EF_NODRAW to renderamt 0 (so they are sent to the client but not rendered by themselves), then change them to renderamt 1 with the current cvar?

chinese-soup:
> Maybe something like change EF_NODRAW to renderamt 0 (so they are sent to the client but not rendered by themselves), then change them to renderamt 1 with the current cvar?

Sounds like a great idea, this means I can get rid of "marking" them as well, will try this.

SmileyAG:
> > Maybe something like change EF_NODRAW to renderamt 0 (so they are sent to the client but not rendered by themselves), then change them to renderamt 1 with the current cvar?
> 
> Sounds like a great idea, this means I can get rid of "marking" them as well, will try this. (Although this of course still means it's doing server-side edits to the demos, meaning entities that wouldn't normally be in the packets will be there, just "fully transparent")

So?

For segmented runs, it is great that it will be hidden in demos so that it does not appear in the render of the run.

chinese-soup:
Done the way @YaLTeR suggested, this should make the most sense:
If you use bxt_show_hidden_entities_clientside 1 and you are making a demo (segmented run), this will make the game add it to the demo, but when you record those runs you can then use bxt_show_hidden_entities_clientside 0 so they don't actually draw during the playback. This should basically be what was originally requested. :)

**Here's a quick vid, first showing in-game with hidden entities and then showing the demo playback:**
https://user-images.githubusercontent.com/5108747/145724084-ae8aa797-f33a-4522-8a26-3bfef24dece4.mp4

YaLTeR (approved):

YaLTeR:
Looks good

YaLTeR:
Thanks!

PR #233, opened 2021-12-13
hud_origin: additional mode for draw positions via last_vieworg method

SmileyAG:
Allows the player to know the position of the camera while using `bxt_freecam 1`.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -510,9 +510,19 @@ namespace CustomHud
 			std::ostringstream out;
 			out.setf(std::ios::fixed);
 			out.precision(precision);
-			out << "X: " << cl.last_vieworg[0] << "\n"
-				<< "Y: " << cl.last_vieworg[1] << "\n"
-				<< "Z: " << cl.last_vieworg[2];
+
+			if (CVars::bxt_hud_origin.GetInt() == 2)
+			{
+				out << "X: " << cl.last_vieworg[0] << "\n"
+					<< "Y: " << cl.last_vieworg[1] << "\n"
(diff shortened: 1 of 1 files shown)


YaLTeR:
I think `hud_origin` should be left with `player.origin` since that's the real player entity origin which is important in certain cases, whereas `cl.last_vieworg` is affected by all kinds of weird bobbing and whatnot IIRC. What's your use-case for checking the camera origin?

SmileyAG:
> What's your use-case for checking the camera origin?

For teleport to the same coordinates at which the camera was in freecam mode.

Maybe it would be better then to make an additional mode which will use `cl.last_vieworg` method, like `bxt_hud_origin 2`?

YaLTeR:
> Maybe it would be better then to make an additional mode which will use cl.last_vieworg method, like bxt_hud_origin 2?

Eh I guess that works, sure.

BunnymodXT/hud_custom.cpp
```diff
@@ -510,9 +510,18 @@ namespace CustomHud
...
+				out << "X: " << player.origin[0] << "\n"
+					<< "Y: " << player.origin[1] << "\n"
+					<< "Z: " << player.origin[2];
+			}
+
+			if (CVars::bxt_hud_origin.GetInt() == 2) {
+				out << "X: " << cl.last_vieworg[0] << "\n"
+					<< "Y: " << cl.last_vieworg[1] << "\n"
+					<< "Z: " << cl.last_vieworg[2];
+			}
```
YaLTeR, line 524: if GetInt == 2 { ... } else { original behavior }

YaLTeR:
Thanks

PR #236, opened 2021-12-15
Added a HUD for draw armor in float value

SmileyAG:
Matherunner hldoc:

изображение

Useful in segmenting when u want to know how much armor you will get from pickup a battery or using a suit charger.

Half-Life HUD can draw that you had a 15 armor, but in truly you can have a 16 armor in the game (if the value is close to 16, e.g. 15.99)


YaLTeR:
Thanks

PR #237, opened 2021-12-15
Use GL_POINTS method for dot instead of rectangle if bxt_cross was 2

SmileyAG:


YaLTeR:
Why?

SmileyAG:
> Why?

I dislike when dot becomes a rectangle tbh, so let's player give a chance to draw it as a real point lmao.

YaLTeR:
Could you attach some screenshots pls?

SmileyAG:
> Could you attach some screenshots pls?

bxt_cross 1:

изображение

bxt_cross 2:

изображение

YaLTeR:
I think it would make more sense to add a new variable like bxt_cross_dot_circle for this.

SmileyAG:
> I think it would make more sense to add a new variable like bxt_cross_dot_circle for this.

I'm not sure if its works for all other users as well. Can you confirm that it works on like your specifications? (Linux, AMD GPU)

If it will be works, then I will make a separate cvar like `bxt_cross_dot_circle`.

On my NVIDIA GPU (GTX 1060) & Windows 10, `GL_POINTS` is displayed as a point, but I have a suspicion that this may not be the case everywhere, since you mentioned earlier that` GL_POINTS` can be rendered as a square?

YaLTeR:
This is what I get with bxt_cross 2 from an artifact from this PR

image

YaLTeR:
Well still call it circle, maybe someone else implements it differently eventually

SmileyAG:
> Well still call it circle, maybe someone else implements it differently eventually

Hoboken said me that it be rendered as point in fullscreen mode, but be a square in windowed 🤔.

YaLTeR:
Then perhaps not worth it, and need to be implemented in another way?

PR #238, opened 2021-12-15
Added bxt_show_player_in_hltv for render player when watching a POV demo with dem_forcehltv 1

SmileyAG:
Borrowed from https://github.com/advancedfx/advancedfx/blob/main/AfxHookGoldSrc/HltvFix.cpp


YaLTeR:
`bxt_show_player_in_hltv`?

YaLTeR:
Thanks

PR #241, opened 2021-12-21
Added bxt_ch_get_pos

chinese-soup:
inspired by getpos in hl2games:

```
] getpos
setpos -7771.200195 -10095.799805 961.250000;setang 1.583999 41.712093 0.000000
```

didnt use any fancy stringstream stuff, cuz dont think its needed tbh, this is just a quick and dirty "getpos and use it in 2 minutes bruh" type beat cmd

EDIT: walled behind "ch" because bxt_hud_origin is not whitelisted either, get_angles could be allowed outside "ch" since it'S whitelisted, but i think its better to have it consistent for both

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -2108,6 +2108,21 @@ struct HwDLL::Cmd_BXT_CH_Set_Armor
 	}
 };
 
+struct HwDLL::Cmd_BXT_CH_Get_Origin_And_Angles
+{
+	NO_USAGE();
+
+	static void handler()
+	{
+		auto &hw = HwDLL::GetInstance();
+		float angles[3];
+		hw.GetViewangles(angles);
+		hw.ORIG_Con_Printf("bxt_ch_set_angles %f %f %f;", angles[0], angles[1], angles[2]);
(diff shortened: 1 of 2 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -2108,6 +2108,31 @@ struct HwDLL::Cmd_BXT_CH_Set_Armor
...
 
+struct HwDLL::Cmd_BXT_CH_Get_Origin
+{
+	NO_USAGE();
+
+	static void handler()
+	{
+		auto &hw = HwDLL::GetInstance();
+		hw.ORIG_Con_Printf("bxt_ch_set_pos %f %f %f\n", (*hw.sv_player)->v.origin[0], (*hw.sv_player)->v.origin[1],
+														(*hw.sv_player)->v.origin[2]);
```
chinese-soup, line 2119: image 
image

YaLTeR:
I think it makes sense to just have 1 command for this like in hl2

BunnymodXT/modules/HwDLL.cpp
YaLTeR: whose bright idea at jetbrains it was to use tabs for alignment

chinese-soup:
> I think it makes sense to just have 1 command for this like in hl2

alright

YaLTeR (approved):

YaLTeR:
Thanks

PR #245, opened 2021-12-26
Added SDL=true for bshift_cutsceneless (same DLLs as bshift)

chinese-soup:
Only hooked on steam, therefore doesn't break any existing older mods. 

It's valve DLLs so just like the original game ("bshift") returns true for `BIsValveGame` it has to return true for `BIsValveGame` (aka useLibSDLInput=true) here as well otherwise mouse movement 360° doesn't work. This isn't fixed by RInput (based on Muty's test stream, where until he renamed the folder to "dod" he couldn't get mouse 360° in the mod with rinput or without.)

cc @hobokenn

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -1680,7 +1680,6 @@ void HwDLL::FindStuff()
 		GET_FUTURE(CL_Record_f)
 		GET_FUTURE(build_number);
 		GET_FUTURE(Key_Event);
-		GET_FUTURE(BIsValveGame);
 		#undef GET_FUTURE
 
 		{
@@ -1730,6 +1729,7 @@ void HwDLL::FindStuff()
 		GET_FUTURE(R_DrawWorld);
 		GET_FUTURE(R_DrawEntitiesOnList);
 		GET_FUTURE(R_DrawParticles);
+		GET_FUTURE(BIsValveGame);
 
 		if (oldEngine) {
 			GET_FUTURE(LoadAndDecryptHwDLL);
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -1675,6 +1680,7 @@ void HwDLL::FindStuff()
 		GET_FUTURE(CL_Record_f)
 		GET_FUTURE(build_number);
 		GET_FUTURE(Key_Event);
+		GET_FUTURE(BIsValveGame);
```
hobokenn, line 1683: use other GET_FUTURE macro, otherwise Cbuf_Execute would be nullptr if the function is not found

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5372,3 +5378,10 @@ HOOK_DEF_0(HwDLL, void, __cdecl, R_DrawParticles)
 	ORIG_R_DrawParticles();
 }
 
+HOOK_DEF_1(HwDLL, int, __fastcall, BIsValveGame, void*, thisptr)
```
hobokenn, line 5381: __cdecl with no arguments

BunnymodXT/patterns.hpp
```diff
@@ -446,6 +446,11 @@ namespace patterns
 			"HL-SteamPipe",
 			"55 8B EC 51 A1 ?? ?? ?? ?? 85 C0 75"
 		);
+
+		PATTERNS(BIsValveGame,
+			"HL-SteamPipe",
+			"56 33 F6 8B 86 ?? ?? ?? ??"
```
hobokenn, line 452: no trailing masked bytes

chinese-soup: Thanks, done.

chinese-soup:
Guess I have to switch to `BUsesSDLInput`, calling ORIG_BIsValveGame crashed the game...

YaLTeR:
> Also do you think I should somehow "cache" the pfnGetGameDirectory() result or nah?

Is it called often?

chinese-soup:
> > Also do you think I should somehow "cache" the pfnGetGameDirectory() result or nah?
> 
> Is it called often?

Every frame (fps_max)

YaLTeR:
Maybe we can compute `bool return true from uses sdl input` somewhere once?

YaLTeR:
Thanks

PR #249, opened 2021-12-29
Added bxt_viewmodel_semitransparent

SmileyAG:
Based on code from bunnymod-pro: https://github.com/YaLTeR/bunnymod-pro/blob/master/cl_dll/StudioModelRenderer.cpp#L1558-L1566

`bxt_viewmodel_semitransparent 1`:

изображение


YaLTeR:
`bxt_viewmodel_opacity` reads like something that controls the... opacity of the viewmodel. So at 0.5 it would be half transparent and at 0 it would be completely invisible.

SmileyAG:
> `bxt_viewmodel_opacity` reads like something that controls the... opacity of the viewmodel. So at 0.5 it would be half transparent and at 0 it would be completely invisible.

Hmm, maybe it's better rename it to like `bxt_viewmodel_half_opacity` and change boolean priority in functions? (0/1 instead of 1/0)

YaLTeR:
Why is this needed actually?

SmileyAG:
> Why is this needed actually?

If someone's want a transparent models without modifying weapon .mdl, same reason as we had a `bxt_viewmodel_disable_` cvars.

YaLTeR:
> If someone's want a transparent models

Does someone want?

SmileyAG:
> > If someone's want a transparent models
> 
> Does someone want?

`¯\_(ツ)_/¯`

SmileyAG:
@YaLTeR Can be it merged now?

YaLTeR:
Alright, just call it `bxt_viewmodel_semitransparent` maybe

YaLTeR:
Thanks

PR #254, opened 2022-01-09
Added bxt_force_fov from bxt-rs

SmileyAG:
Used `GetFloat() >= 1.0` instead of `GetBool()`, bcz with <1.0 values FOV not be changed, but in same time player get a semi-wallhack state, so I decided clamp min. value to `1.0`.


YaLTeR:
Thanks

YaLTeR:
It crashes linux hl with a weird backtrace. Commenting out the R_SetFrustum hook prevents the crash.

PR #255, opened 2022-01-09
Rename bxt_ch_set_angles to bxt_set_angles to not ruin future segmented runs via sv_cheats enabled

SmileyAG:
Makes possible to avoid of usage `sv_cheats 1` in segmented runs at all.

This command has non-cheat protected counterparts such as `cl_yawspeed` and `cl_pitchspeed`, but for these cmds u have to experiment with the values to get the correct camera position, so it easier to use `bxt_set_angles` for turn.


YaLTeR:
Thanks

PR #256, opened 2022-01-11
Fix bxt_force_fov on Linux

chinese-soup:
* just missed getting the symbols on linux

* also only register the cvar when found otherwise bxt_force_fov would crash on windows as well if scr_fov_value was nullptr on some version of the game


YaLTeR:
Thanks

PR #261, opened 2022-02-05
Added bxt_fix_mouse_horizontal_limit for enabling SDL mouse fixes from engine

SmileyAG:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5338,7 +5339,7 @@ HOOK_DEF_3(HwDLL, int, __cdecl, DispatchDirectUserMsg, char*, pszName, int, iSiz
 {
 	const char *gameDir = ClientDLL::GetInstance().pEngfuncs->pfnGetGameDirectory();
 
-	if (!std::strcmp(gameDir, "czeror") && !std::strcmp(pszName, "InitHUD"))
+	if ((!std::strcmp(gameDir, "czeror") || !std::strcmp(gameDir, "czeror_cutsceneless")) && !std::strcmp(pszName, "InitHUD"))
```
YaLTeR, line 5342: Unrelated change

YaLTeR (changes requested):
This is certainly not the right way to fix the problem. If anything it should be called `bxt_fix_something` (not sure what exactly it fixes) so the user can set it to 1, while still keeping the manual list of game dirs so the user doesn't need to set it on their own in most cases.

BunnymodXT/modules/HwDLL.cpp
SmileyAG: Yeah I know. I'm just doing a cutsceneless mod for this game, so I decided quickly push all changes as single commit cuz I'm kinda laziness for splitting, but I can do it if needed.

YaLTeR: Would be good yeah.

SmileyAG:
> This is certainly not the right way to fix the problem. If anything it should be called `bxt_fix_something` (not sure what exactly it fixes) so the user can set it to 1, while still keeping the manual list of game dirs so the user doesn't need to set it on their own in most cases.

You mean smth like: `if (ClientDLL::GetInstance().DoesGameDirMatch("bshift_cutsceneless") || CVars::bxt_fix_sdl.GetBool())`, right?

YaLTeR:
Yeah. But with an actual descriptive cvar name.

BunnymodXT/cvars.cpp
```diff
@@ -33,6 +33,7 @@ namespace CVars
 	CVarWrapper bxt_viewmodel_semitransparent("bxt_viewmodel_semitransparent", "0");
 	CVarWrapper bxt_clear_green("bxt_clear_green", "0");
 	CVarWrapper bxt_force_fov("bxt_force_fov", "0");
+	CVarWrapper bxt_fix_mouse_sdl("bxt_fix_mouse_sdl", "0");
```
YaLTeR, line 36: People don't know what sdl is and "fix mouse" doesn't tell me what it actually fixes. What does it actually fix?

SmileyAG: Mouse will be limited in moving horizontal without SDL fix.

https://user-images.githubusercontent.com/58108407/152642909-637f4a38-f04e-47f8-b05a-f5b42b53a565.mp4

YaLTeR: fix_mouse_horizontal_limit?

YaLTeR:
Thanks

PR #263, opened 2022-02-19
hud_custom: add additional _entity_info mode to draw only index/class/target/targetname/hp

SmileyAG:
изображение

I decided to make second mode for `bxt_hud_entity_info`, which will not draw extra options for entities, such as `velocity`, `origin`, `angles` cuz they are not much needed tbh.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -848,7 +848,7 @@ namespace CustomHud
 		if (CVars::bxt_hud_visible_landmarks.GetBool())
 		{
 			int x, y;
-			GetPosition(CVars::bxt_hud_visible_landmarks_offset, CVars::bxt_hud_visible_landmarks_anchor, &x, &y, -20, 0);
+			GetPosition(CVars::bxt_hud_visible_landmarks_offset, CVars::bxt_hud_visible_landmarks_anchor, &x, &y, -75, 0);
 
 			std::ostringstream out;
 			out << "Visible Landmarks:\n";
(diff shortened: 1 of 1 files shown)


YaLTeR:
Eh, not sure. Usually when it goes up from 1 to 2 the amount of information increases.

SmileyAG:
> Eh, not sure. Usually when it goes up from 1 to 2 the amount of information increases.

Done.

BunnymodXT/hud_custom.cpp
```diff
@@ -734,22 +734,30 @@ namespace CustomHud
 				const char *classname = sv.GetString(ent->v.classname);
 				out << classname << '\n';
 
+				if (ent->v.target != 0) {
+					const char *target = sv.GetString(ent->v.target);
+					out << "T: " << target << '\n';
```
YaLTeR, line 739: Not very obvious what "T" and "TN" means...

BunnymodXT/hud_custom.cpp
```diff
@@ -838,7 +846,7 @@ namespace CustomHud
 		if (CVars::bxt_hud_visible_landmarks.GetBool())
 		{
 			int x, y;
-			GetPosition(CVars::bxt_hud_visible_landmarks_offset, CVars::bxt_hud_visible_landmarks_anchor, &x, &y, -20, 0);
+			GetPosition(CVars::bxt_hud_visible_landmarks_offset, CVars::bxt_hud_visible_landmarks_anchor, &x, &y, -75, 0);
```
YaLTeR, line 849: Maybe this also needs to be dependent on iCharHeight or smth?

BunnymodXT/hud_custom.cpp
SmileyAG: I purposely shortened it to fit as many characters as possible from entity names, but I can write it fully if u want.

YaLTeR: Maybe "Target" and "Name"

YaLTeR:
Thanks

PR #264, opened 2022-02-25
Apply removing of skybox to bxt_wallhack as default

SmileyAG:


YaLTeR:
I would be for this, but I'd ask if anyone is using bxt_skybox_remove just in case.

SmileyAG:
> I would be for this, but I'd ask if anyone is using bxt_skybox_remove just in case.

Well, maybe it would be better to not remove cmd then, and instead still make a commit about applying of removing skybox for wallhack as default?

YaLTeR:
Well, did you ask?

SmileyAG:
> Well, did you ask?

I didn't asked anyone. I just realized that some things can be seen with removed skybox so thats why I'm want to keep it:

https://user-images.githubusercontent.com/58108407/157143867-45f92744-2a81-4432-beef-5c59fdd0cb3a.mp4

YaLTeR:
Ah, okay

YaLTeR:
Thanks

PR #267, opened 2022-03-07
Opposing Force nightvision sprite toggling

chinese-soup:
QoL: currently runners replace the nightvision sprite with an empty transparent one to get this effect.

image
image


YaLTeR:
Thanks

PR #268, opened 2022-03-07
Don't stop timer on crash

chinese-soup:
As per Kananga's and @mxpph idea/request:

Kananga hosts HL1 races and he uses people's in-game timer on their respective streams (because of stream lag etc.) for comparing the final time to see "who won". During the last HL1 race he hosted a person crashed on hornet (bxt prevented) and Kananga wants to give people who crash a chance, so he had to start his timer again, but he did that late and they had to manually add the time to his timer, hence the suggested idea of a racemode was born. 

I'd link a timestamp'd video, but he didn't highlight it atm.

The run is still invalid, but people in the race will run with `bxt_racemode_on_crash 1` in order to not lose the timer for Kananga to use.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -25,6 +25,7 @@ namespace CustomHud
 	static bool receivedAccurateInfo = false;
 	static playerinfo player;
 	bool countingTime = false;
+	bool invalidRun = false;
 	int hours = 0, minutes = 0, seconds = 0;
 	double timeRemainder = 0.0;
 	int frames = 0;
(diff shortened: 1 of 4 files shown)


YaLTeR:
I was thinking maybe instead of a separate cvar, just keep the timer going on crash, but color it red or something, and still with the chat message?

chinese-soup:
* Changed behavior as per suggestion, got rid of the cvar.
* Added "invalid run" message under timer as well (yes, some people might have the timer in the bottom corner of the screen, but at that point, those runners probably know that they've crashed), since I feel like it's more descriptive than just a red timer.
* Set bxt_hud_timer to 1 if it's set to 0 and the run crashed, motivation is people can have hud_saytext set to 0 and may not know they've crashed. (Of course, the mod will still know from demos, since it's a server side say as mentioned in the OG PR).

Behavior showed off here:
https://user-images.githubusercontent.com/5108747/157266085-cb68067c-4ec2-4e1a-bd33-96214d242cfd.mp4

1.  Start timer + load crash save -> invalid run
2. Stop timer -> still invalid run
3. Reset timer -> invalid run flag is reset
4. Load crash save with timer not started (0.00) -> no invalid run flag, since run isn't happening
5. Lastly show off bxt_hud_timer setting to 1

PR #269, opened 2022-03-08
Fix refdef (bxt_freecam, etc.) for Op4 Linux

chinese-soup:
The problem with Op4 LInux is that for some reason `V_CalcRefDef` is never called (even though it's found) and both Ghidra (and halfflife-op4-updated) doesn't suggest why it shouldn't be called. `V_CalcNormalRefDef` is called but I can't find from where, I don't wanna hook `V_CalcNormalRefdef` cuz mods can modify top-level `V_CalcRefdef` (and keeping it to V_CalcRefDef for other mods would require stoopid if statements)

So uhhh, a little sketchy, but `V_CalcRefDef` has to be found every time anyway (the game crashes if ClientDLL.cpp can't find original `V_CalcRefDef`), so basically what I'm doing here is instead of this engine func that calls the ClientDLL's V_Calcrefdef:
```c++
void ClientDLL_CalcRefdef(ref_params_s *pparams)

{
  if (cl_funcs.pCalcRefdef != (HUD_CALCREF_FUNC *)0x0) {
                    /* WARNING: Could not recover jumptable at 0x001c04f9. Too many branches */
                    /* WARNING: Treating indirect jump as call */
    (*cl_funcs.pCalcRefdef)();
    return;
  }
  return;
}
```
I am calling the ClientDLL's already-hooked `V_CalcRefDef` function, this fixes it for Op4 and doesn't break it for other mods because engine would literally just call the ClientDLL's `V_CalcRefdef` anyway.

For frusaha 
image

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -286,6 +286,11 @@ extern "C" void __cdecl R_SetFrustum()
 {
 	HwDLL::HOOKED_R_SetFrustum();
 }
+
+extern "C" void __cdecl ClientDLL_CalcRefdef(ref_params_s *pparams)
+{
+	ClientDLL::HOOKED_V_CalcRefdef(pparams);
+}
 #endif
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks

PR #271, opened 2022-03-11
Added patterns for Aliens vs Predator 2 modification

SmileyAG:


YaLTeR:
Thanks! For the next time: offsets should be added in the same commit as the pattern, because having just the pattern without the offsets will lead to an `assert(false)`. The general idea of commit splitting is to be able to checkout any commit and have it build and work correctly. This is important for searching which commit introduced breakage.

PR #272, opened 2022-03-16
Added bxt_hud_game_color and bxt_hud_game_alpha

SmileyAG:
Allow to change in-game HUD color and alpha.

# Confirmation from @chinese-soup that it works on Linux:

# HL1 (255 alpha, "255 100 255" rgb):

изображение

# OP4 (255 alpha, "0 255 255" rgb):

изображение

# CSCZDS ("255 255 255" rgb):

изображение

### DIFF
--- a/BunnymodXT/modules/ClientDLL.cpp
+++ b/BunnymodXT/modules/ClientDLL.cpp
@@ -1086,16 +1086,13 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, HUD_Reset)
 
 HOOK_DEF_2(ClientDLL, void, __cdecl, HUD_Redraw, float, time, int, intermission)
 {
-	if (!CVars::bxt_hud_game_color.IsEmpty()) {
+	custom_hud_color_set = false;
+
+	if (sscanf(CVars::bxt_hud_game_color.GetString().c_str(), "%hhu %hhu %hhu", &custom_r, &custom_g, &custom_b) == 3)
 		custom_hud_color_set = true;
-		std::istringstream ss(CVars::bxt_hud_game_color.GetString());
-		ss >> custom_r >> custom_g >> custom_b;
-	}
 
 	if (!CVars::bxt_disable_hud.GetBool())
 		ORIG_HUD_Redraw(time, intermission);
-
-	custom_hud_color_set = false;
 
 	CustomHud::Draw(time);
 }
@@ -1407,8 +1404,8 @@ HOOK_DEF_4(ClientDLL, void, __cdecl, ScaleColors, int*, r, int*, g, int*, b, int
 		*b = custom_b;
 	}
 
-	if (CVars::bxt_hud_game_alpha.GetInt() >= 1 && CVars::bxt_hud_game_alpha.GetInt() <= 255)
-		a = CVars::bxt_hud_game_alpha.GetInt();
+	if (CVars::bxt_hud_game_alpha.GetFloat() > 0 && CVars::bxt_hud_game_alpha.GetFloat() <= 1)
+		a = static_cast<int>(CVars::bxt_hud_game_alpha.GetFloat() * 255);
 
 	ORIG_ScaleColors(r, g, b, a);
 }
--- a/BunnymodXT/modules/ClientDLL.cpp
+++ b/BunnymodXT/modules/ClientDLL.cpp
@@ -1404,8 +1404,8 @@ HOOK_DEF_4(ClientDLL, void, __cdecl, ScaleColors, int*, r, int*, g, int*, b, int
 		*b = custom_b;
 	}
 
-	if (CVars::bxt_hud_game_alpha.GetFloat() > 0 && CVars::bxt_hud_game_alpha.GetFloat() <= 1)
-		a = static_cast<int>(CVars::bxt_hud_game_alpha.GetFloat() * 255);
+	if (CVars::bxt_hud_game_alpha.GetInt() >= 1 && CVars::bxt_hud_game_alpha.GetInt() <= 255)
+		a = CVars::bxt_hud_game_alpha.GetInt();
 
 	ORIG_ScaleColors(r, g, b, a);
 }
--- a/BunnymodXT/modules/ClientDLL.cpp
+++ b/BunnymodXT/modules/ClientDLL.cpp
@@ -1086,8 +1086,16 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, HUD_Reset)
 
 HOOK_DEF_2(ClientDLL, void, __cdecl, HUD_Redraw, float, time, int, intermission)
 {
+	if (!CVars::bxt_hud_game_color.IsEmpty()) {
+		custom_hud_color_set = true;
+		std::istringstream ss(CVars::bxt_hud_game_color.GetString());
+		ss >> custom_r >> custom_g >> custom_b;
+	}
+
 	if (!CVars::bxt_disable_hud.GetBool())
 		ORIG_HUD_Redraw(time, intermission);
+
+	custom_hud_color_set = false;
 
 	CustomHud::Draw(time);
 }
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1361,3 +1390,21 @@ HOOK_DEF_1(ClientDLL, void, __cdecl, CHudFlashlight__drawNightVision_Linux, void
 		return;
 	ORIG_CHudFlashlight__drawNightVision_Linux(thisptr);
 }
+
+HOOK_DEF_4(ClientDLL, void, __cdecl, ScaleColors, int*, r, int*, g, int*, b, int, a)
+{
+	if (!CVars::bxt_hud_game_color.IsEmpty()) {
+		unsigned custom_r = 0, custom_g = 0, custom_b = 0;
+		std::istringstream ss(CVars::bxt_hud_game_color.GetString());
+		ss >> custom_r >> custom_g >> custom_b;
```
YaLTeR, line 1399: Bad idea to parse cvar value here, this is called many times every frame.

YaLTeR:
So as far as I can see this (as expected) leaves some colors unchanged, and I'm guessing it also overwrites some different colors to be the same color. Is this something that people will want to use in this state?

BunnymodXT/modules/ClientDLL.cpp
SmileyAG: > Bad idea to parse cvar value here, this is called many times every frame.

Should I do similar as like in `DrawCrosshair`?

https://github.com/YaLTeR/BunnymodXT/blob/89083a08c748b99e8c7b124d7f1f985146141a0b/BunnymodXT/hud_custom.cpp#L1272-L1284

YaLTeR: It's kind of the same issue there tbh, but it's worse with ScaleColors because it's called many times per frame rather than just once.

SmileyAG:
> So as far as I can see this (as expected) leaves some colors unchanged, and I'm guessing it also overwrites some different colors to be the same color. Is this something that people will want to use in this state?

Default hardcoded HUD colors are returned even after disabling `bxt_hud_game_color`, so there is no reason for worries that colors can be overwritten:

https://user-images.githubusercontent.com/58108407/158748804-84537ab2-2000-4d24-92ed-6c55b2c24551.mp4

YaLTeR:
I mean like in this video, low health is supposed to be red, but it becomes the overwritten color.

SmileyAG:
> I mean like in this video, low health is supposed to be red, but it becomes the overwritten color.

Well, we can make a separate cmd that will checks the player's HP and if it's <= 25 value, then the custom colors will be declined, smth like: `bxt_hud_game_color_red_on_low_health 1`

If you are worries that some of HL1 mods may change the HP behavior for red color, we can change the priority from boolean to integer for that command, and then the player can himself set the amount of hp at which the HUD would be red, like: `bxt_hud_game_color_red_on_low_health 25`

YaLTeR:
Nah, let's not do that

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1057,9 +1086,17 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, HUD_Reset)
 
 HOOK_DEF_2(ClientDLL, void, __cdecl, HUD_Redraw, float, time, int, intermission)
 {
+	if (!CVars::bxt_hud_game_color.IsEmpty()) {
+		custom_hud_color_set = true;
+		std::istringstream ss(CVars::bxt_hud_game_color.GetString());
+		ss >> custom_r >> custom_g >> custom_b;
```
YaLTeR, line 1092: It should only set `custom_hud_color_set` if the parsing succeeded. Actually probably will be easier with `sscanf` (which returns the number of arguments parsed successfully). Also `sscanf` probably will avoid an allocation?

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1057,9 +1086,17 @@ HOOK_DEF_0(ClientDLL, void, __cdecl, HUD_Reset)
...
+	}
+
 	if (!CVars::bxt_disable_hud.GetBool())
 		ORIG_HUD_Redraw(time, intermission);
 
+	custom_hud_color_set = false;
```
YaLTeR, line 1098: Put this above the first if.

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1361,3 +1398,17 @@ HOOK_DEF_1(ClientDLL, void, __cdecl, CHudFlashlight__drawNightVision_Linux, void
...
+
+HOOK_DEF_4(ClientDLL, void, __cdecl, ScaleColors, int*, r, int*, g, int*, b, int, a)
+{
+	if (custom_hud_color_set) {
+		*r = custom_r;
+		*g = custom_g;
+		*b = custom_b;
+	}
+
+	if (CVars::bxt_hud_game_alpha.GetInt() >= 1 && CVars::bxt_hud_game_alpha.GetInt() <= 255)
```
YaLTeR, line 1410: Maybe this should be a float, and `a` should be multiplied by it? This way if you set _alpha 0.5 then the HUD will be half as opaque, while preserving the different alpha values within the HUD. Will need to clamp the final a between 0 and 255 though.

BunnymodXT/modules/ClientDLL.hpp
```diff
@@ -63,6 +64,10 @@ class ClientDLL : public IHookableNameFilter
 
 	bool DoesGameDirMatch(const char *game);
 
+	unsigned custom_r = 0, custom_g = 0, custom_b = 0;
+
```
YaLTeR, line 68:
```suggestion
```

BunnymodXT/modules/ClientDLL.cpp
SmileyAG: Like that?

```
unsigned char custom_r, custom_g, custom_b; // hpp file

HOOK_DEF_2(ClientDLL, void, __cdecl, HUD_Redraw, float, time, int, intermission)
{
	custom_hud_color_set = false;

	if (sscanf(CVars::bxt_hud_game_color.GetString().c_str(), "%hhu %hhu %hhu", &custom_r, &custom_g, &custom_b) == 3) {
		custom_hud_color_set = true;
	}
	...
}
```

BunnymodXT/modules/ClientDLL.cpp
SmileyAG: > Maybe this should be a float, and `a` should be multiplied by it? This way if you set _alpha 0.5 then the HUD will be half as opaque, while preserving the different alpha values within the HUD. Will need to clamp the final a between 0 and 255 though.

Maybe I will saying dumb question, but how I get it without warning about data loss? Cuz that what I got:

```
if (CVars::bxt_hud_game_alpha.GetFloat() > 0 && CVars::bxt_hud_game_alpha.GetFloat() <= 1)
		a = CVars::bxt_hud_game_alpha.GetFloat() * 255;
```

BunnymodXT/modules/ClientDLL.cpp
YaLTeR: Yeah something like that.

BunnymodXT/modules/ClientDLL.cpp
YaLTeR: `a = static_cast<int>(CVars::bxt_hud_game_alpha.GetFloat() * a);`

and also a = min(255, max(0, a))

YaLTeR: and >= 0 rather than > 0

SmileyAG: > and >= 0 rather than > 0

I want to cancel custom overriding of alpha if cvar was == 0, if u want full transparent HUD - better to disable them tbh.

YaLTeR: Ah, fair

YaLTeR: I still suggest multiplying `a` though

SmileyAG: > I still suggest multiplying `a` though

What's the point of this? I meant like I clamped some values by cvar, so it will multiply to 255 for values that higher > 0 and <= 1, so there shouldn't be some issues with that.

YaLTeR: so if there's some hud element with alpha 100 and other with alpha 200 and you set _alpha 0.5, they become alpha 50 and 100, so one is still two times more transparent than the other

YaLTeR:
Thanks

PR #276, opened 2022-03-23
entity_info: allow to see the exact origin of the brush entities

SmileyAG:
Old behaviour:

изображение

New behaviour:

изображение

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -780,7 +780,7 @@ namespace CustomHud
 					// Borrowed from https://github.com/SNMetamorph/goldsrc-monitor/blob/08c368e246d09996b2d85e4367d4d8cc1e507712/sources/library/displaymode_entityreport.cpp#L45
 					Vector origin;
 
-					if (strstr(classname, "func_") != NULL)
+					if (ent->v.solid == SOLID_BSP || ent->v.movetype == MOVETYPE_PUSHSTEP)
 						origin = ent->v.origin + ((ent->v.mins + ent->v.maxs) / 2.f);
 					else
 						origin = ent->v.origin;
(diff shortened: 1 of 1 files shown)


BunnymodXT/hud_custom.cpp
```diff
@@ -773,13 +773,16 @@ namespace CustomHud
 					}
 				}
 
+				// Borrowed from https://github.com/SNMetamorph/goldsrc-monitor/blob/08c368e246d09996b2d85e4367d4d8cc1e507712/sources/library/displaymode_entityreport.cpp#L45-L48
+				const Vector origin = ent->v.origin + ((ent->v.mins + ent->v.maxs) / 2.f);
```
YaLTeR, line 777: I think this should only be done for brush entities. For entities with an actual origin it should show the origin as is.

BunnymodXT/hud_custom.cpp
```diff
@@ -777,9 +777,17 @@ namespace CustomHud
 				{
 					out << "Yaw: " << ent->v.angles[1] << '\n';
 
-					out << "X: " << ent->v.origin.x << '\n';
-					out << "Y: " << ent->v.origin.y << '\n';
-					out << "Z: " << ent->v.origin.z << '\n';
+					// Borrowed from https://github.com/SNMetamorph/goldsrc-monitor/blob/08c368e246d09996b2d85e4367d4d8cc1e507712/sources/library/displaymode_entityreport.cpp#L45
+					Vector origin;
+
+					if (strstr(classname, "func_") != NULL)
```
YaLTeR, line 783: this is a pretty cursed way to check for this

SmileyAG: How I should do check that, then? According to TWHL, most of brush entities is `func_` or `trigger_`: https://twhl.info/wiki/page/category%3ABrush_Entities

SmileyAG: @YaLTeR Oh, I think if I'll be check for `ent->v.solid == SOLID_BSP`, would be that works?

YaLTeR:
Thanks

Issue #141, opened 2021-02-06
bxt_hud_useables 1 cause crashing in Counter-Strike: Condition Zero Deleted Scenes


SmileyAG:

SmileyAG:
That what VS Debugger shows me.
![unknown](https://user-images.githubusercontent.com/58108407/107257552-3aba0a00-6a54-11eb-9f28-63ecf3f83b00.png)

YaLTeR:
Can't pinpoint the issue from this screenshot alone. Someone would have to look at this more in depth.

hobokenn:
it is crashing because the offset for `ObjectCaps` is 0x18 in CZDS

SmileyAG:
> 
> 
> it is crashing because the offset for `ObjectCaps` is 0x18 in CZDS

Can you make pull request with that changes?

hobokenn:
> 
> 
> > it is crashing because the offset for `ObjectCaps` is 0x18 in CZDS
> 
> Can you make pull request with that changes?

i wouldn't know how to do that without breaking the current offset, the current one is also wrong for spirit based mods

SmileyAG:
I should manually close this issue, since it got fixed in #155 (https://github.com/YaLTeR/BunnymodXT/pull/155/commits/37e8c53acb9292d4c7834e1b432da8d84d3833b5#diff-7e7d27826cbffc76caea454a313ac0c866da8271fbc1b4b508ab2af9aef9464eR560)

PR #277, opened 2022-03-31
entity_info: check if door has a flag for usable

SmileyAG:
That spawnflag prevent door from opening by touch, but it allow for `+use` or being triggered.

https://user-images.githubusercontent.com/58108407/161095705-2c30f047-2762-43ef-8f81-339f72f2ddcf.mp4

The advantage over `bxt_hud_useables 1` for that case that in different mods, incorrect `offFuncObjectCaps` might crash game as like it been in #141 issue.


YaLTeR:
Thanks

PR #278, opened 2022-04-04
Added bxt_render_far_entities

### LINKED ISSUE #225: Make bxt_novis not affect gameplay
Is it possible to remove PVS for player and not let allow that for NPCs somehow?

Not want to brainwash with HLSDK & engine code for find workaround way... so, @YaLTeR, some thoughts?

```
<-- — 15.11.2021
yeah, it affects (shows) entities that would normally be outside of PVS and some of those could behave differently when outside of it 
dunno any particular cases myself, but i guess quad does))
quadkid — 15.11.2021
yeah, the ninja run patterns was different in my RP TAS when I tried 3rd person shots with novis
<-- — 16.11.2021
yea antoher PVS example is the op4 RTA trick ppl do these days
(aka if you had novis 1 for that you wouldnt be able to do it)
since the dying soldier would be loaded
```

SmileyAG:
Closes #225 

About `kRenderFxEntInPVS` from Spirit SDK: https://github.com/LogicAndTrick/halflife-updated/blob/mods/twhl-tower-2/dlls/client.cpp#L1274

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -208,6 +208,7 @@ void ServerDLL::Clear()
 	offBhopcap = 0;
 	pBhopcapWindows = 0;
 	pCZDS_Velocity_Byte = 0;
+	pAddToFullPack_PVS_Byte = 0;
 	offm_iClientFOV = 0;
 	offm_rgAmmoLast = 0;
 	maxAmmoSlots = 0;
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1768,6 +1794,19 @@ HOOK_DEF_7(ServerDLL, int, __cdecl, AddToFullPack, struct entity_state_s*, state
 	auto oldRenderAmount = ent->v.renderamt;
 	auto oldRenderFx = ent->v.renderfx;
 
+	if (CVars::bxt_novis_serverside.GetBool())
+	{
+		if (pAddToFullPack_PVS_Byte)
+		{
+			MemUtils::ReplaceBytes(reinterpret_cast<void*>(pAddToFullPack_PVS_Byte), 1, reinterpret_cast<const byte*>(CVars::bxt_novis_serverside.GetBool() ? "\xEB" : "\x74"));
```
YaLTeR, line 1801: This removes the check but it never adds it back when the cvar is disabled.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1768,6 +1794,19 @@ HOOK_DEF_7(ServerDLL, int, __cdecl, AddToFullPack, struct entity_state_s*, state
 	auto oldRenderAmount = ent->v.renderamt;
 	auto oldRenderFx = ent->v.renderfx;
 
+	if (CVars::bxt_novis_serverside.GetBool())
```
YaLTeR, line 1797: Unclear name. `bxt_novis` comes from `r_novis` or something (and it is also serverside). Maybe this can be more descriptive like `bxt_render_far_entities`?

YaLTeR (changes requested):

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: Oh, right. I wrote that logic being a too sleepy, and for some reason added a extra check which makes no return byte back LMAO

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1768,6 +1794,12 @@ HOOK_DEF_7(ServerDLL, int, __cdecl, AddToFullPack, struct entity_state_s*, state
 	auto oldRenderAmount = ent->v.renderamt;
 	auto oldRenderFx = ent->v.renderfx;
 
+	if (pAddToFullPack_PVS_Byte)
```
YaLTeR, line 1797: Could you also check for the existing byte value for extra safety, like in bxt_bhopcap iirc?

YaLTeR:
Thanks!

PR #284, opened 2022-04-16
bxt_ch_get_pos shows a origin of player camera if bxt_hud_origin is 2

SmileyAG:


YaLTeR:
I guess this is because bxt_hud_origin shows camera pos at 2?

PR #285, opened 2022-04-30
DispatchDirectUserMsg: return 0 in pszName for demo playback

SmileyAG:


YaLTeR:
What's this for?

SmileyAG:
> What's this for?

1) Added demo API header from HLSDK to make easier in future make some actions which is should be applied for demo playback and nothing more
2) Might be it would be for a small make perfomance better since it doesn't call while you don't playback demo?

YaLTeR:
> Might be it would be make perfomance better for a little cuz it doesn't call until you don't playback demo?

Uhh huh? I don't think it's worth it for this. Also the correct way would be to add the condition to the start so it's actually checked before the others (and potentially skipping the following checks).

SmileyAG:
Merge?

YaLTeR:
> Uhh huh? I don't think it's worth it for this.

SmileyAG:
> 

When I first time made that demo crash fix I been more less aware about HLSDK coding and stuff, so that's the reason why I didn't made that pDemoAPI check from initial.

Like for that time that I segmenting CSCZDS I didn't noticed issues with that fix out of demo playback (like in-game), but I'm still scare for some behaviour at case so better to add it for full safety than not having it.

SmileyAG:
Ahhh, nevermind. I decided to check `hw.so` again and found that`InitHUD` in `DispatchDirectUserMsg` calls only when you using `playdemo` or `viewdemo` commands. So that's pull request can be closed, sorry for taking your time :)

PR #286, opened 2022-05-03
Refactoring of call exported functions to make less dependant from client & server

SmileyAG:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -890,6 +894,12 @@ void HwDLL::FindStuff()
 		else
 			EngineDevWarning("[hw dll] Could not find CBaseUI::HideGameUI [Linux].\n");
 
+		ServerDLL::GetInstance().ppGlobals = reinterpret_cast<globalvars_t*>(MemUtils::GetSymbolAddress(m_Handle, "gGlobalVariables"));
```
YaLTeR, line 897: ServerDLL stuff should be found in ServerDLL not in HwDLL

YaLTeR: If these two pointers make more sense in HwDLL then they should be moved into HwDLL I think

SmileyAG: > ServerDLL stuff should be found in ServerDLL not in HwDLL

I don't think that makes sense since `ppGlobals` at now been found at HwDLL for Windows & Linux.

And If I'm going to move some pointer which is find in `HwDLL` to `ServerDLL`, that would make it looks wrong in debug console, like:

```
[server dll] ....
[hw dll] gGlobalVariables
[server dll] ....
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -1216,15 +1216,15 @@ void HwDLL::FindStuff()
 				default:
 				case 0: // HL-Steampipe
 					ServerDLL::GetInstance().pEngfuncs = *reinterpret_cast<enginefuncs_t**>(reinterpret_cast<uintptr_t>(LoadThisDll) + 95);
```
YaLTeR, line 1218: I'd say server engfuncs should also move to HwDLL then and the same applies to the client ones? Not sure, how would that look?

SmileyAG:
@YaLTeR Merge?

YaLTeR:
Seems to work but there are two new warnings that should be fixed: https://github.com/YaLTeR/BunnymodXT/runs/6318451504?check_suite_focus=true#step:7:48

SmileyAG:
> Seems to work but there are two new warnings that should be fixed: [YaLTeR/BunnymodXT/runs/6318451504?check_suite_focus=true#step:7:48](https://github.com/YaLTeR/BunnymodXT/runs/6318451504?check_suite_focus=true#step:7:48)

Done.

YaLTeR:
Thanks

PR #287, opened 2022-05-07
Added engine method for changing HUD color

SmileyAG:
Pros:

- Allows to change HUD color in mods which is not used `ScaleColors` (Like: Gunman Chronicles)
- Doesn't require support for custom DLLs

Cons:

- Alpha is hardcoded to 255 in `SPR_Set` function, and it's not stored in some variable for easy change.
That means that it requires to reverse-engineering and rewrite entire function to remove that hardcode shit. If we would do this, then we can completely removing `ScaleColors` from BXT.


ScriptedSnark:
Tested on Linux.

Without bxt_hud_game_engineside:
 without
 
 With bxt_hud_game_engineside:
 with

SmileyAG:
> Tested on Linux.
> 
> Without bxt_hud_game_engineside: without
> 
> With bxt_hud_game_engineside: with

Thanks for test! Yeah that should be like that cuz with `bxt_hud_game_engineside` HUD starts to been hardcoded with 255 alpha as I said before.

SmileyAG:
@YaLTeR Review now.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5665,3 +5716,39 @@ HOOK_DEF_0(HwDLL, void, __cdecl, R_SetFrustum)
...
+		g = cl.custom_g;
+		b = cl.custom_b;
+	}
+
+	ORIG_SPR_Set(hSprite, r, g, b);
+}
+
+HOOK_DEF_2(HwDLL, void, __cdecl, DrawCrosshair, int, x, int, y)
+{
+	insideDrawCrosshair = true;
```
YaLTeR, line 5736: What's this for?

YaLTeR:
Thanks

PR #289, opened 2022-05-08
Added bxt_disable_loading_text

SmileyAG:
I think that useful when you want to make raw capturing of TAS, cuz as I seen there is been 1 frame of loading in it, but if the text would be disabled - then it would be not even noticed at all.

https://youtu.be/TD62NRRfe8w?t=77 (1:17)


YaLTeR:
Sounds like a better fit for bxt-rs then?

YaLTeR:
Added to bxt-rs.

PR #290, opened 2022-05-09
Fix freezing of startup vids in pre-Steampipe builds with BXT_DISABLEPROCESSWINGHOSTING environment variable

SmileyAG:


YaLTeR:
What is this and why?

> Window ghosting is a Windows Manager feature that lets the user minimize, move, or close the main window of an application that is not responding.

Sounds like something we don't want to disable, especially not for... startup vids?..

SmileyAG:
> What is this and why?

That fixes freezing of playback startup vids after first 5 seconds. I copied that solution from TWHL Discord:

solo

I'm pretty sure there is some GoldSrc runners who might be want to use that feature for some meme-modding game, cuz why not `¯\_(ツ)_/¯`

SmileyAG:
I forgot to tell months ago that freeze thing is also disappears if some of debugger attached to **hl.exe** process.

So I wonder is it possible to use some functions from Microsoft API to load some like Windows debugger and then de-attach it after video playback process done? Pretty sure it should be possible to make.

YaLTeR:
Sounds cursed. Definitely not something for BXT just for startup vids.

PR #293, opened 2022-05-19
Log player.health value to demos

SmileyAG:


SmileyAG:
Merge? @YaLTeR 

Reason why I added `bxt_hud_health_override_in_demo`, because at the moment there is no changes for parsing health data from demo to in-game HUD

And, even it would be done later in which I hope, I can make suspicions that it would be works correctly only for `playdemo`, but not for `viewdemo` (cuz you can set any timestamp in it without watching frames at which health changes)

YaLTeR:
Why does this remove bxt_show_player_in_hltv?

SmileyAG:
> Why does this remove bxt_show_player_in_hltv?

I don't think it's needs? Or at least I could return it back, but set to 1 as default.

YaLTeR:
Like, regardless, it's completely unrelated and should be a separate PR.

SmileyAG:
> Like, regardless, it's completely unrelated and should be a separate PR.

Can be merged now.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5181,9 +5186,18 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
 {
 	static double timeCounter = 0.0;
 	static bool usePassedTime = false;
-
```
YaLTeR, line 5184: Unrelated change.

BunnymodXT/modules/ServerDLL.hpp
```diff
@@ -78,6 +79,8 @@ class ServerDLL : public IHookableDirFilter
 
 	entvars_t *obboPushable = nullptr;
 
+	int m_afHealthLast;
```
YaLTeR, line 82:
```suggestion
	int lastRecordedHealth;
```
BXT doesn't use the `m_af` notation.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5041,6 +5041,11 @@ void HwDLL::SaveInitialDataToDemo()
 		RuntimeData::Add(RuntimeData::Edicts{ maxEdicts });
 	}
 
+	auto &hw = HwDLL::GetInstance();
+	int playerhealth = static_cast<int>((*hw.sv_player)->v.health);
+
+	RuntimeData::Add(RuntimeData::PlayerHealth{playerhealth});
+
```
YaLTeR, line 5048: This should set `lastRecordedHealth`.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5181,9 +5186,18 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
 	auto minFrametime = CVars::_bxt_min_frametime.GetFloat();
 
+	auto &hw = HwDLL::GetInstance();
+
+	if (IsRecordingDemo())
+	{
+		int playerhealth = static_cast<int>((*hw.sv_player)->v.health);
+
+		if (playerhealth != ServerDLL::GetInstance().m_afHealthLast)
+			RuntimeData::Add(RuntimeData::PlayerHealth{playerhealth});
```
YaLTeR, line 5198: This should also set `lastRecordedHealth`.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2161,3 +2173,15 @@ HOOK_DEF_3(ServerDLL, void, __fastcall, CChangeLevel__TouchChangeLevel, void*, t
...
 	return ORIG_CChangeLevel__TouchChangeLevel(thisptr, edx, pOther);
 }
+
+HOOK_DEF_1(ServerDLL, void, __cdecl, PlayerPostThink, edict_t*, pEntity)
+{
+	if (HwDLL::GetInstance().IsRecordingDemo())
+	{
+		entvars_t *pev = &pEntity->v;
+
+		m_afHealthLast = static_cast<int>(pev->health);
```
YaLTeR, line 2183: This way there's no need to hook this function because the health is already updated in the place where it matters.

BunnymodXT/runtime_data.hpp
```diff
@@ -52,6 +52,10 @@ namespace RuntimeData
 		int edicts;
 	};
 
+	struct PlayerHealth {
+		int playerhealth;
```
YaLTeR, line 56:
```suggestion
		int health;
```

YaLTeR (changes requested):

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: Better to have that function, since in next PRs I going to make also some stuff like `if (oldflags & FL_ONGROUND)` for checking player flags on previous frame in `hud_custom` module for `bxt_hud_keys`.

YaLTeR: No, I did not mean that as a "you can avoid hooking a function if you set lastRecordedHealth in those other places", I meant it as a "it would be more correct to set lastRecordedHealth in those other places, so please do that, and when you do that, this hooks becomes empty, so you can remove it". If it will be needed for a future PR, please add it in the future PR, not here.

SmileyAG:
@YaLTeR Now?

BunnymodXT/cvars.cpp
```diff
@@ -128,6 +128,7 @@ namespace CVars
 	CVarWrapper bxt_hud_health("bxt_hud_health", "0");
 	CVarWrapper bxt_hud_health_offset("bxt_hud_health_offset", "");
 	CVarWrapper bxt_hud_health_anchor("bxt_hud_health_anchor", "0.5 1");
+	CVarWrapper bxt_hud_health_override_in_demo("bxt_hud_health_override_in_demo", "0");
```
YaLTeR, line 131: The recording to demo part is good to merge, however this cvar I don't really like (passing the value through a cvar shouldn't be required here). Since it's not hooked up anyway (nothing sets it), maybe you could remove it from the PR?

SmileyAG: https://github.com/YaLTeR/BunnymodXT/blob/eff47ffdc2e8485733d15c87a72b316b6be08fe3/BunnymodXT/hud_custom.cpp#L1016-L1019

That should be a temporary solution for renders, before someone would implement the proper parsing of health from BXT demo data to `bxt_hud_health` (#294), so I don't want to remove that cvar at the moment.

SmileyAG: Because at the moment I could dump health from BXT Demo Data with special tool and then set that value to `bxt_hud_health_override_in_demo` command, so yeah, that useful.

YaLTeR: Alright I suppose

PR #295, opened 2022-05-23
Always render a local player model when watching a POV demo with dem_forcehltv 1

SmileyAG:
Don't mind about failed builds, it would be fixed after you first merging #293 request and then this one with merge-rebase button.


YaLTeR:
Could you rebase so the CI runs?

YaLTeR:
Thanks

PR #296, opened 2022-05-25
ammohistory: replaced engine method for custom colors to clientside for hud fading

SmileyAG:
Generally, engine method for changing colors is a nicely than clientside, but when it comes to HUD fading, it is bad - since alpha is being forced to 255 in `SPR_Set`.

Ammohistory is single HUD element which makes notice that issue, all other HUDs is fine to change at engineside.


YaLTeR:
Thanks

PR #304, opened 2022-06-07
Extended Draw_FillRGBA pattern for HL-WON-1712

SmileyAG:
At steampipe dll, there's been a some other unrelated function with same byte sequence, so that's reason why I extended


YaLTeR:
Thanks

PR #306, opened 2022-06-07
Added cl_righthand

### LINKED ISSUE #219: cl_righthand

SmileyAG:
Closes #219

# Screenshot

изображение

# Note

On Windows, `IsFlippedViewModel` function been called at **R_GLStudioDrawPoints** itself:

изображение

But, linux compilers of .so binaries somehow turn that function to few checks instead, so that a main reason why I can't rename cvar:

изображение


YaLTeR:
Maybe just call the cvar `cl_righthand`? Then it will work automatically.

YaLTeR:
¯\_(ツ)_/¯

SmileyAG:
@YaLTeR Review now.

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -714,6 +714,13 @@ void ClientDLL::RegisterCVarsAndCommands()
 	EngineDevMsg("[client dll] Registering CVars.\n");
 
 	#define REG(cvar) HwDLL::GetInstance().RegisterCVar(CVars::cvar)
+
+	if (!HwDLL::GetInstance().ORIG_Cvar_FindVar(CVars::cl_righthand.GetPointer()->name))
```
YaLTeR, line 718:
```suggestion
	if (!HwDLL::GetInstance().ORIG_Cvar_FindVar("cl_righthand"))
```
Simpler is better in this case I think

SmileyAG: Done.

YaLTeR:
Thanks

PR #307, opened 2022-06-11
Don't apply HLTV model fix if dem_forcehltv command not found, fix cl_righthand check not working

SmileyAG:
https://github.com/YaLTeR/BunnymodXT/blob/0e492eed05ee48cb4d2e911380ef948cdd05b315/BunnymodXT/modules/ClientDLL.cpp#L1330

That makes prevent crashing of 1712 build when it tries to find `IsSpectateOnly`, since it's not exist in that build as `dem_forcehltv` cmd as well.


BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1327,7 +1327,7 @@ HOOK_DEF_3(ClientDLL, int, __cdecl, HUD_AddEntity, int, type, cl_entity_s*, ent,
 
 HOOK_DEF_0(ClientDLL, int, __cdecl, CL_IsThirdPerson)
 {
-	if (pEngfuncs->pDemoAPI->IsPlayingback() && pEngfuncs->IsSpectateOnly())
+	if (pEngfuncs->pDemoAPI->IsPlayingback() && HwDLL::GetInstance().ORIG_Cmd_FindCmd("dem_forcehltv") && pEngfuncs->IsSpectateOnly())
```
YaLTeR, line 1330: This will do a string search on all commands every time this function runs. This has nontrivial performance overhead, better to do it some other way (perhaps searching for the command once at startup).

YaLTeR (changes requested):

BunnymodXT/modules/ClientDLL.cpp
SmileyAG: I did a lot of researching and found that best place for that checks would be `HUD_Init` function, since it calls only once when ClientDLL is loaded.

I also moved `cl_righthand` check to `HUD_Init` too, since I didn't noticed that it isn't worked before :KEKW

SmileyAG:
`cl_righthand` makes no sense to move at `HUD_Frame`, since initialize in engine comes with `HUD_Init`:

SmileyAG:
After a dozen of force-pushing, I finally got a good working checks now, so I will not touch this PR anymore, unless there is would be some requsted review for changes.

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1140,6 +1140,9 @@ HOOK_DEF_1(ClientDLL, void, __cdecl, HUD_Frame, double, time)
 {
 	ORIG_HUD_Frame(time);
 
+	if (!orig_forcehltv_found && HwDLL::GetInstance().ORIG_Cmd_FindCmd("dem_forcehltv"))
```
YaLTeR, line 1143: If the command is missing, this check will still run every frame I'm afraid.

SmileyAG: I think check should be looks like that then:

```
if (!stop_call_forcehltv_per_frame && HwDLL::GetInstance().ORIG_Cmd_FindCmd("dem_forcehltv")) {
  orig_forcehltv_found = true;
  stop_call_forcehltv_per_frame = true;
}
else if (!stop_call_forcehltv_per_frame && !HwDLL::GetInstance().ORIG_Cmd_FindCmd("dem_forcehltv")) {
  stop_call_forcehltv_per_frame = true;
}
```

YaLTeR: Could be simpler:

```
if (check_forcehltv) {
  check_forcehltv = false;
  found = FindCmd("dem_forcehltv");
}
```

YaLTeR:
Thanks

PR #308, opened 2022-06-13
Added bxt_disable_conback_in_demo for WON engine builds

SmileyAG:
This command is extremely needed since I am going to re-render some old segmented run on that build.

`console 0` is disabling only characters, but not background itself, so that the reason why I did that command.

I know that porting to bxt-rs would a better solution, but there is no hooked `cl_enginefuncs` table in it, which means I cannot get easy access to the DemoAPI.


SmileyAG:
Oh, hoboken said that there is actually hooked `cls_demos` in **bxt-rs**, from which you get access to **cls_demos.demoplayback**, so I'll might be port it later to bxt-rs, going to close that request here then.

PR #311, opened 2022-06-14
Fixed sounds missing while demo playback on old WON builds

### LINKED ISSUE #309: Injected BunnymodXT causes sounds missing while demo playback on old WON builds
Like footsteps or ladder climb for example, I don't know what a issue causes that.

But if I'll launch normal game or inject bxt-rs only - then it not appears.

SmileyAG:
Closes #309


YaLTeR:
Big thonk tbh

PR #314, opened 2022-06-16
Added bxt_force_clear

SmileyAG:
`gl_clear 1` acts so weird (even w/o BXT), cuz sometimes it can clear to **red** color e.g. (when I expected **black**), so that's reason why I make that cvar.

It's been mentioned years ago by shar:


YaLTeR:
Thanks

PR #315, opened 2022-06-17
health_drawdamage: replaced engine method for custom colors to clientside for hud fading

SmileyAG:


YaLTeR:
Still battling with HUD colors I see

PR #317, opened 2022-07-01
Fixed bxt_force_jumpless not working on ladders

SmileyAG:
See: https://github.com/ValveSoftware/halflife/blob/master/pm_shared/pm_shared.c#L2129-L2133

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -291,6 +291,7 @@ void ServerDLL::FindStuff()
 			offPunchangles = 0xA0;
 			offWaterlevel = 0xE4;
 			offInDuck = 0x90;
+			offFlags = 0xB8;
 			offBasevelocity = 0x74;
 		});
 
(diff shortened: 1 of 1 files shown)


SmileyAG:
> Could you go through the game and check that it works everywhere now?

I didn't passed a full game with that changes, but from quick tests I can say that everything should fine to work.

Because I got analyzed all `IN_JUMP` matches in HLSDK code, and all of that changes being fit, explanation:

1) Removing call of `PM_Jump` makes to remove jumping sound
2) Forcing `*flags |= FL_WATERJUMP` to disallow player get speed boost from jumping on conveyor belts
3) Forcing `cmd->buttons &= ~IN_JUMP` to disallow player jumping from the ladder

SmileyAG:
@YaLTeR Merge now? I don't think that it can be broke smth, check for `FL_WATERJUMP` flag in HLSDK is used only in **CBasePlayer::Jump** function and nowhere else.

`cmd->buttons &= ~IN_JUMP` disallows to jumping. (and it resets themself after disabling cvar, which is good cuz not requires extra code)

`return;` at the beginning of **PM_Jump** function completely disables jumping code & their sounds.

YaLTeR:
Alright cool

PR #322, opened 2022-07-20
Necromancing CS1.6 support from cs branch

khanghugo:
From https://github.com/YaLTeR/BunnymodXT/compare/master...cs

This was a feature long time ago that enables further TAS capability in CS1.6 but long remains unsupported. I just want it back.

One thing, Windows pattern is 7 years old but I've checked the instructions for client.dll and it looks reasonable enough to work.

Here are the bytes to patch (just using cl_yawspeed as an example)
pattern1
pattern2

And this is the result
pattern_patched
The patch will push cl_yawspeed and cl_pitchspeed value onto top stack regardless of any branching involving comparison between itself and the two other limit values. 

I then TAS it a bit and it works. The prediction closely matches the simulator, or it just is. Before, the yaw lock would not only prevent the script from strafing closer but also all the new fancy camera smoothing stuffs.

For attributions, I have never written a line of code in my entire life so most of things here are simply mirroring already existing code and mainly the one you had initially in the cs branch.


BunnymodXT/patterns.hpp
```diff
@@ -909,6 +909,26 @@ namespace patterns
...
+
+		PATTERNS(CS_AngleSpeedCap_Linux,
+			"CS-SteamPipe",
+			"57 56 53 83 EC 30 F6 05 ?? ?? ?? ?? 01"
+		);
+
+		PATTERNS(CS_SpeedScaling,
+			"CS-SteamPipe",
+			"D9 FA DD DB DD D8 DD D8 D8 54 24 18 DF E0 25 00 41 00 00 75 1C D8 7C 24 18 D9 C0 D8 4E 10 D9 5E 10 D9 C0 D8 4E 14 D9 5E 14 D8 4E 18 D9 5E 18 EB 02"
+			);
```
YaLTeR, line 926:
```suggestion
		);
```

BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -946,6 +998,61 @@ bool ClientDLL::DoesGameDirMatch(const char *game)
...
 }
 
+void ClientDLL::SetAngleSpeedCap(bool capped) {
+	if (!pCS_AngleSpeedCap && !pCS_AngleSpeedCap_Linux) {
+		return;
+	}
+
+	if (capped) { // restore the bytes
+		if (pCS_AngleSpeedCap) {
+			MemUtils::ReplaceBytes(reinterpret_cast<void*>(pCS_AngleSpeedCap + 5), 1, reinterpret_cast<const byte*>("\x7B"));
```
YaLTeR, line 1008: Since this and SetSpeedScaling are run every frame, and replacing bytes involves potentially performance-sensitive operations like changing the memory protection flags, before calling `ReplaceBytes` you should check if the bytes need to be replaced. I believe the `bxt_bhopcap` code does that as an example.

YaLTeR (changes requested):
Instead of just updating the HLStrafe submodule commit to your fork, you should PR the change into HLStrafe, and when that's merged, subsequently update the submodule to point at the commit in the main HLStrafe repo.

khanghugo:
Thanks for the feedback. I didn't really intend to have these commits flooding over here. Let me see what I can do to clean them up.

YaLTeR (approved):

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4734,6 +4736,14 @@ HLStrafe::MovementVars HwDLL::GetMovementVars()
 	vars.Stepsize = CVars::sv_stepsize.GetFloat();
 	vars.Bounce = CVars::sv_bounce.GetFloat();
 	vars.Bhopcap = CVars::bxt_bhopcap.GetBool();
+	vars.BhopcapScale = 0.65f;
+	vars.SpeedScale = 1.7f;
+
+	if (ClientDLL::GetInstance().DoesGameDirMatch("cstrike")) {
+		vars.Maxspeed = 250.0f; // sv_maxspeed does not override the speed unless below 300
```
SmileyAG, line 4743: Can you remove that maxspeed hack and go check #325 request?

SmileyAG (changes requested):

BunnymodXT/modules/HwDLL.cpp
khanghugo: Thanks for the help. I got it working perfectly with that.

BunnymodXT/hud_custom.cpp
```diff
@@ -1429,6 +1429,21 @@ namespace CustomHud
 		}
 	}
 
+	void DrawStamina(float flTime)
+	{
+		if (CVars::bxt_hud_stamina.GetBool())
+		{
+			int x, y;
+			GetPosition(CVars::bxt_hud_stamina_offset, CVars::bxt_hud_stamina_anchor, &x, &y, -200, (si.iCharHeight * 16) + 4);
```
SmileyAG, line 1437: Need to fix offset here, since that overlaps entity info HUD: 
изображение

SmileyAG (changes requested):

BunnymodXT/hud_custom.cpp
khanghugo: This is at relative offset 29, right above self gauss.
image

khanghugo: image
This is another one in the middle, barely occupied by visible landmark hud. I think it should be okay there.

khanghugo:
At this stage, it is all there is. I have yet tested on Windows so I am not really sure how it would be. For further testing, I can make a small TAS for verification.

khanghugo:
At this stage, it seems pretty good. Not sure how well it would work on Windows but all the stuffs are there.

khanghugo:
Here is a small TAS that can be used to verified on Windows. Made with the commit.

https://youtu.be/vh2AdTZyWoQ

khanghugo:
Beside that, lots of things are still borked. Eg: you cannot record demo or use other things like hud_useable. It needs some more work.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4735,6 +4738,20 @@ HLStrafe::MovementVars HwDLL::GetMovementVars()
 	vars.Bounce = CVars::sv_bounce.GetFloat();
 	vars.Bhopcap = CVars::bxt_bhopcap.GetBool();
 
+	if (cl.DoesGameDirMatch("cstrike")) {
```
YaLTeR, line 4741: Please cache this with a `static bool is_cstrike = cl.DoesGameDirMatch("cstrike");` so that we don't have a string comparison every frame. Also I guess when the maxspeed PR is merged (pending on a similar change) this should be rebased on top of that.

YaLTeR:
You should be able to rebase on top of master now to get Smiley's client maxspeed thing.

khanghugo:
Yep, rebase is here. If there is anything else, I can fix.

YaLTeR:
Thanks! Finally, the CS 1.6 support is back!

PR #323, opened 2022-07-24
Added bxt_disable_gamedir_check_in_demo

SmileyAG:
# Example of use-case:

HL verification moderators or usual runners want to watch demos which is been recorded on `valve` gamedir, while launching game from `valve_WON` (and vice-versa, cuz NGHL runners usually drops WON dlls to `valve` folder, while runners on HL2005 or GSP 2.3 packages playing from `valve_WON`).


YaLTeR:
Thanks

PR #324, opened 2022-07-26
bxt_hud_weapon_cooldown

Eddio0141:
`bxt_hud_weapon_cooldown` to show weapon cooldown information.

I'm still trying to figure out 2 things
- I don't know what the offset of [this](https://github.com/Eddio0141/BunnymodXT/blob/c7a200a61cb3747db8138587cf9faef0bb503c3d/BunnymodXT/modules/ServerDLL.cpp#L2323) is about, or if there is a better way to getting player's active item.
- Some weapons in half life echoes work but some doesn't such as the shotgun. Weapons spawned by `impulse 101` works fine. Still haven't figured out a way to fix this.


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2178,6 +2245,105 @@ bool ServerDLL::GetGonarchInfo(float &health, int& sequence, float& frame) const
...
+	auto activeItemBase = *reinterpret_cast<uintptr_t*>(pActiveItem + 0x4);
+	if (!activeItemBase)
+		return;
+	// TODO what is this offset?
+	auto activeItem = reinterpret_cast<edict_t*>(activeItemBase - 0x80);
+	if (!activeItem || !pEngfuncs->pfnEntOffsetOfPEntity(activeItem))
+		return;
+
+	auto* gauss = pEngfuncs->pfnFindEntityByString(nullptr, "classname", "weapon_gauss");
+	if (gauss && pEngfuncs->pfnEntOffsetOfPEntity(gauss) && !std::strcmp(HwDLL::GetInstance().ppGlobals->pStringBase + activeItem->v.classname, "weapon_gauss")) {
```
YaLTeR, line 2333: Hm, what if there's a `weapon_gauss` on the floor but the player also has a `weapon_gauss`?

BunnymodXT/hud_custom.cpp
```diff
@@ -1009,6 +1009,105 @@ namespace CustomHud
...
+	{
+		return std::get<1>(l) < std::get<1>(r);
+	}
+
+	void DrawWeaponCooldownInfo(float flTime)
+	{
+		if (CVars::bxt_hud_weapon_cooldown.GetBool())
+		{
+			int x, y;
+			GetPosition(CVars::bxt_hud_weapon_cooldown_offset, CVars::bxt_hud_weapon_cooldown_anchor, &x, &y, 0, si.iCharHeight + 3);
```
YaLTeR, line 1022: Hm, maybe we should draw it at the bottom left, above the HP? (avoiding the damage indicators somehow) Because right now at 1280x720 you can already reasonably get a TAS editor status overlap:

image

Eddio0141: yeah seems better, ill move it down there

Eddio0141:
Made it so its right below the timer since the cooldown hud can grow downwards

PR #325, opened 2022-08-09
TAS-related: Get client player maxspeed in HL modifications if it's available

SmileyAG:
About factor hack in PARANOIA: https://github.com/FWGS/paranoia_toolkit/blob/b5a073543499f593e8c12bce27840614a5f47d32/code_src/pm_shared/pm_shared.c#L1216-L1217

### DIFF
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -11,7 +11,6 @@ namespace CVars
 	CVarWrapper _bxt_bunnysplit_time_update_frequency("_bxt_bunnysplit_time_update_frequency", "41");
 	CVarWrapper _bxt_save_runtime_data_in_demos("_bxt_save_runtime_data_in_demos", "1");
 	CVarWrapper _bxt_tas_script_generation("_bxt_tas_script_generation", "1337");
-	CVarWrapper bxt_tas_clientmaxspeed_in_prediction("bxt_tas_clientmaxspeed_in_prediction", "1");
 	CVarWrapper bxt_taslog_filename("bxt_taslog_filename", "taslogger.log");
 	CVarWrapper bxt_autopause("bxt_autopause", "0");
 	CVarWrapper bxt_interprocess_enable("bxt_interprocess_enable", "0");
@@ -187,7 +186,6 @@ namespace CVars
 		&_bxt_norefresh,
 		&_bxt_save_runtime_data_in_demos,
 		&_bxt_tas_script_generation,
-		&bxt_tas_clientmaxspeed_in_prediction,
 		&bxt_taslog_filename,
 		&bxt_autopause,
 		&bxt_interprocess_enable,
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -11,6 +11,7 @@ namespace CVars
 	CVarWrapper _bxt_bunnysplit_time_update_frequency("_bxt_bunnysplit_time_update_frequency", "41");
 	CVarWrapper _bxt_save_runtime_data_in_demos("_bxt_save_runtime_data_in_demos", "1");
 	CVarWrapper _bxt_tas_script_generation("_bxt_tas_script_generation", "1337");
+	CVarWrapper bxt_tas_clientmaxspeed_in_prediction("bxt_tas_clientmaxspeed_in_prediction", "1");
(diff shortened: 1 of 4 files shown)


BunnymodXT/cvars.cpp
```diff
@@ -11,6 +11,7 @@ namespace CVars
 	CVarWrapper _bxt_bunnysplit_time_update_frequency("_bxt_bunnysplit_time_update_frequency", "41");
 	CVarWrapper _bxt_save_runtime_data_in_demos("_bxt_save_runtime_data_in_demos", "1");
 	CVarWrapper _bxt_tas_script_generation("_bxt_tas_script_generation", "1337");
+	CVarWrapper bxt_tas_clientmaxspeed_in_prediction("bxt_tas_clientmaxspeed_in_prediction", "1");
```
YaLTeR, line 14: Why is this needed?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4733,11 +4746,19 @@ void HwDLL::FindCVarsIfNeeded()
...
 	auto vars = HLStrafe::MovementVars();
+	auto &cl = ClientDLL::GetInstance();
 
 	FindCVarsIfNeeded();
 	vars.Frametime = GetFrameTime();
 	vars.Maxvelocity = CVars::sv_maxvelocity.GetFloat();
-	vars.Maxspeed = CVars::sv_maxspeed.GetFloat();
+
+	if (CVars::bxt_tas_clientmaxspeed_in_prediction.GetBool() && cl.DoesGameDirMatch("paranoia"))
+		vars.Maxspeed = cl.pEngfuncs->GetClientMaxspeed() * 3.2f; // GetMaxSpeed is factor here, 3.2f is approx. multiplier for walkmove
```
YaLTeR, line 4756: Shouldn't it be multiplied by sv_maxspeed?

SmileyAG: > Shouldn't it be multiplied by sv_maxspeed?

Got checked rn, and you are right, then formula for getting maxspeed in **PARANOIA** would be like:

`cl.pEngfuncs->GetClientMaxspeed() * CVars::sv_maxspeed.GetFloat() / 100.0f`

BunnymodXT/cvars.cpp
SmileyAG: Well, with hardcoding clientmaxspeed prediction enabling by default, there is would be no ability to set lower custom maxspeed value in CS games e.g.

Let's say that USP is gives `250` units in CS, but I want to limit it for `200` for some weird reason, then:

- I do `sv_maxspeed 200` and it set lower maxspeed
- But, `GetClientMaxspeed` still goes to output the original value of weapon maxspeed

YaLTeR: Then it should take minimum between sv_maxspeed and client maxspeed?

SmileyAG: > Then it should take minimum between sv_maxspeed and client maxspeed?

I think u could do more simpler by adding `CVars::sv_maxspeed.GetFloat() > cl.pEngfuncs->GetClientMaxspeed()` condition here: 

https://github.com/YaLTeR/BunnymodXT/blob/6396f62e30088e40ecf06fccb659ea6719223b66/BunnymodXT/modules/HwDLL.cpp#L4757

YaLTeR: You should double check what actually happens in the movement code for different client maxspeed vs. sv_maxspeed

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4733,11 +4745,19 @@ void HwDLL::FindCVarsIfNeeded()
...
 {
 	auto vars = HLStrafe::MovementVars();
+	auto &cl = ClientDLL::GetInstance();
 
 	FindCVarsIfNeeded();
 	vars.Frametime = GetFrameTime();
 	vars.Maxvelocity = CVars::sv_maxvelocity.GetFloat();
-	vars.Maxspeed = CVars::sv_maxspeed.GetFloat();
+
+	if (cl.DoesGameDirMatch("paranoia"))
```
YaLTeR, line 4754: Can we cache this somehow? Like with a `static bool is_paranoia`. Because it's a string comparison every frame, which is expensive.

YaLTeR (commented):
This should be good to merge otherwise right?

YaLTeR:
Thanks

PR #328, opened 2022-08-26
Fixed khanghugo maxspeed usage for CS & CSCZ

SmileyAG:
As I said a few pull requests earlier, if someone would use a custom value for `sv_maxspeed` that lower of a weapon maxspeed, then u would get a wrong results. 

And I fixed it with add `CVars::sv_maxspeed.GetFloat() > cl.pEngfuncs->GetClientMaxspeed()` condition, but khang ignored that fact seems `¯\_(ツ)_/¯`


YaLTeR:
Thanks

khanghugo:
Whoops. Sorry.

PR #329, opened 2022-09-05
Added env variable BXT_DISABLE_DEBUG_CONSOLE

chinese-soup:
Tested on windows 10 with goldsrc package 2.4 adding `SET BXT_DISABLE_DEBUG_CONSOLE=1` to the batch file.

https://user-images.githubusercontent.com/5108747/188501369-b1cd1591-6678-45e8-8724-8c837ed9a424.mp4

### DIFF
--- a/BunnymodXT/Windows/conutils.cpp
+++ b/BunnymodXT/Windows/conutils.cpp
@@ -9,6 +9,10 @@ namespace ConUtils
 
 	void Init()
 	{
+		auto debugConsoleDisabled = std::getenv("BXT_DISABLE_DEBUG_CONSOLE");
+		if (debugConsoleDisabled)
+			return;
+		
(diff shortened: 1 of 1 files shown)


BunnymodXT/Windows/conutils.cpp
```diff
@@ -9,6 +9,10 @@ namespace ConUtils
 
 	void Init()
 	{
+		auto debugConsoleDisabled = std::getenv("BXT_DISABLE_DEBUG_CONSOLE");
+		if(debugConsoleDisabled)
```
YaLTeR, line 13:
```suggestion
		if (debugConsoleDisabled)
```
:)

chinese-soup: ye

PR #331, opened 2022-09-17
Adding AXN style stuffs for CS1.6

khanghugo:
Just a stamina-related cvar to remove its effect on jump height and slow down.

Also bhop cap pattern for CS1.6. Not sure if I am doing the right way. Yet to test on Windows but I think this would work just fine.


khanghugo:
Pattern `D8 88 F4 01 00 00` repeats sometimes for both Windows and Linux binary. I guess I can make this less verbose a bit.

khanghugo:
Whoops, just realize the `pBhopcapWindows` patches different instruction. I don't want to complicate that.

khanghugo:
I have not tested any TAS related things so let's see.

khanghugo:
All works great. Now it is possible to do hlkz map in cs now, or just plain old axn maps

YaLTeR:
Thanks!

PR #332, opened 2022-09-18
Endsection handling

fireblizzard:
Hazard Course ends the game via the _trigger_endsection_ entity, and possibly some other mods or campaigns. This PR just handles this game ending method so that the timer is stopped and LiveSplit stops too, because previously we had to stop both manually when running Hazard Course and it was a bit annoying, specially when you're new to running the thing and don't know what's going on with the timer/splits.

Smiley mentioned that there's some case where sending the player to the menu via _trigger_endsection_ is not meant to end the game, but only one level or something, and then the player goes back to chapter selection or whatever and continues from there. For this reason, it won't automatically end the game on `bxt_timer_autostop 1` unless it's Hazard Course. The player will have to set `bxt_timer_autostop 2` if they actually want the game to end, so they have this workaround and they don't need to wait for someone to hardcode their mod's trigger targetname for it to work out of the box.


YaLTeR:
Thanks

PR #334, opened 2022-09-22
Added bxt_fire_on_button_*

SmileyAG:


BunnymodXT/cvars.cpp
```diff
@@ -73,6 +73,9 @@ namespace CVars
 	CVarWrapper bxt_render_far_entities("bxt_render_far_entities", "0");
 	CVarWrapper bxt_fire_on_mm_targetname("bxt_fire_on_mm_targetname", "");
 	CVarWrapper bxt_fire_on_mm_command("bxt_fire_on_mm_command", "");
+	CVarWrapper bxt_fire_on_button("bxt_fire_on_button", "0");
+	CVarWrapper bxt_fire_on_button_target("bxt_fire_on_button_target", "");
```
YaLTeR, line 77: Why not do this like with multimanager above? So it's one less cvar. Also why filter by target instead of targetname?

SmileyAG: Most of buttons doesn't `targetname`, but they always have a `target` to activate.

YaLTeR:
Thanks

PR #335, opened 2022-09-23
Added bxt_fire_on_stuck, second viewangles mode for take punches into angle account

SmileyAG:
@jorikgrozniy asked me to implement those features, so I did!

### DIFF
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -58,6 +58,8 @@ namespace CVars
 	CVarWrapper skill;
 
 	// Serverside CVars
+	CVarWrapper bxt_stucksave("bxt_stucksave", "0");
+	CVarWrapper bxt_stucksave_command("bxt_stucksave_command", "");
 	CVarWrapper bxt_autojump("bxt_autojump", "0");
 	CVarWrapper bxt_autojump_priority("bxt_autojump_priority", "1");
 	CVarWrapper bxt_bhopcap("bxt_bhopcap", "1");
@@ -71,6 +73,7 @@ namespace CVars
 	CVarWrapper bxt_force_duck("bxt_force_duck", "0");
 	CVarWrapper bxt_force_jumpless("bxt_force_jumpless", "0");
 	CVarWrapper bxt_render_far_entities("bxt_render_far_entities", "0");
+	CVarWrapper bxt_fire_on_mm("bxt_fire_on_mm", "0");
 	CVarWrapper bxt_fire_on_mm_targetname("bxt_fire_on_mm_targetname", "");
 	CVarWrapper bxt_fire_on_mm_command("bxt_fire_on_mm_command", "");
 	CVarWrapper bxt_remove_stamina("bxt_remove_stamina", "0");
@@ -243,6 +246,8 @@ namespace CVars
 		&fps_max,
 		&default_fov,
 		&skill,
+		&bxt_stucksave,
+		&bxt_stucksave_command,
 		&bxt_autojump,
 		&bxt_bhopcap,
 		&bxt_timer_autostop,
@@ -264,6 +269,7 @@ namespace CVars
 		&bxt_wallhack_alpha,
 		&bxt_novis,
 		&bxt_render_far_entities,
+		&bxt_fire_on_mm,
 		&bxt_fire_on_mm_targetname,
 		&bxt_fire_on_mm_command,
 		&bxt_show_triggers,
--- a/BunnymodXT/cvars.cpp
+++ b/BunnymodXT/cvars.cpp
@@ -58,8 +58,7 @@ namespace CVars
 	CVarWrapper skill;
 
 	// Serverside CVars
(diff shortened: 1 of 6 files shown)


BunnymodXT/cvars.cpp
```diff
@@ -71,6 +73,7 @@ namespace CVars
 	CVarWrapper bxt_force_duck("bxt_force_duck", "0");
 	CVarWrapper bxt_force_jumpless("bxt_force_jumpless", "0");
 	CVarWrapper bxt_render_far_entities("bxt_render_far_entities", "0");
+	CVarWrapper bxt_fire_on_mm("bxt_fire_on_mm", "0");
```
YaLTeR, line 76: Unrelated change, and back when you added it we specifically decided the other way

BunnymodXT/hud_custom.cpp
```diff
@@ -557,14 +557,25 @@ namespace CustomHud
...
+
 			int x, y;
 			GetPosition(CVars::bxt_hud_viewangles_offset, CVars::bxt_hud_viewangles_anchor, &x, &y, -200, (si.iCharHeight * 10) + 2);
 
 			std::ostringstream out;
 			out.setf(std::ios::fixed);
 			out.precision(precision);
-			out << "Pitch: " << player.viewangles[0] << "\n"
+
+			if (CVars::bxt_hud_viewangles.GetInt() == 2)
```
YaLTeR, line 569: Could you add some indication when you have `bxt_hud_viewangles 2` so it looks different? Otherwise someone's gonna forget or leave it on by accident and then waste time trying to understand why their HUD is showing incorrect values.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1426,6 +1445,14 @@ void ServerDLL::LogPlayerMove(bool pre, uintptr_t pmove) const
 
 HOOK_DEF_1(ServerDLL, void, __cdecl, PM_PlayerMove, qboolean, server)
 {
+	if (CVars::bxt_stucksave.GetBool() && !CVars::bxt_stucksave_command.IsEmpty() && ORIG_PM_CheckStuck())
```
YaLTeR, line 1448: Why is a separate `bxt_stucksave` needed? Just leave the empty check

YaLTeR (changes requested):

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1426,6 +1445,14 @@ void ServerDLL::LogPlayerMove(bool pre, uintptr_t pmove) const
...
+	{
+		std::ostringstream ss;
+		ss << CVars::bxt_stucksave_command.GetString().c_str() << "\n";
```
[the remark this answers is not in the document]
SmileyAG: Is it better rename to `bxt_fire_on_stuck` or smth like that?

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: Because otherwise it would spamming commands each frame when you got stucked, so I could disable it with like:

`bxt_stucksave_command "s4v3;bxt_stucksave 0"`

BunnymodXT/hud_custom.cpp
SmileyAG: What indication you want like? Change color? Some tag like `Pitch (second mode):`?

YaLTeR: What effects does cl.last_viewangles include compared to player.viewangles?

BunnymodXT/modules/ServerDLL.cpp
YaLTeR: Maybe instead it should only run the command when the player was not stuck last frame? Then you don't need to disable and enable it manually.

BunnymodXT/hud_custom.cpp
SmileyAG: Client and server punchangles:

YaLTeR: What about stuff like viewbobbing?

YaLTeR: If it's just punchangle then smth like Angles (incl. punch)

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1520,6 +1521,8 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, PM_PlayerMove, qboolean, server)
 	}
 
 	CustomHud::UpdatePlayerInfo(velocity, origin);
+
+	not_stuck_prev_frame = stuck_cur_frame ? false : true;
```
YaLTeR, line 1525:
```suggestion
	not_stuck_prev_frame = !stuck_cur_frame;
```

Would've saved some trouble if you made it `stuck_prev_frame` rather than `not_stuck_prev_frame`

YaLTeR:
Thanks

PR #336, opened 2022-09-24
Added bxt_fire_on_mm for disabling mm command fire

SmileyAG:


SmileyAG:
Released that if I want to disable mm feature, I can just type random stuff in `bxt_fire_on_mm_targetname`, so that pull request can be closed

PR #337, opened 2022-09-29
Fixed crash when PM_CheckStuck pattern not found

SmileyAG:
Damn I forgot about that xDDDDD


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1539,7 +1543,8 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, PM_PlayerMove, qboolean, server)
 
 	CustomHud::UpdatePlayerInfo(velocity, origin);
 
-	not_stuck_prev_frame = !stuck_cur_frame;
+	if (ORIG_PM_CheckStuck)
```
YaLTeR, line 1546: This doesn't need the check, also while we're at it, could you move this `not_stuck_prev_frame = !stuck_cur_frame;` line all the way above to right after `if (ORIG_PM_CheckStuck)`?

PR #340, opened 2022-10-05
Adding bxt_hud_jumpdistance

khanghugo:
Pretty primitive stuff given the fact that it is not using lots of things.

Ladder jump will not have green number because of this implementation but the distance is still there. I think the way they do calculate ladder jump distance in KZ is very different.

Mostly copy-pasted code from other function.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -670,83 +670,42 @@ namespace CustomHud
 		if (CVars::bxt_hud_jumpdistance.GetBool())
 		{
 			static bool inJump = false;
-			static float lastTime = flTime;
-			static double passedTime = FADE_DURATION_JUMPSPEED;
-			static int fadingFrom[3] = { hudColor[0], hudColor[1], hudColor[2] };
 			static double jumpDistance = 0.0;
 
-			int r = hudColor[0],
-				g = hudColor[1],
-				b = hudColor[2];
-
-			if (FADE_DURATION_JUMPSPEED > 0.0f)
+			// 1 = jumpdistance will update when velocity is reasonably positive
+			// not 1 = jumpdistance will update when velocity changes, eg just falling
+			if (!inJump && (
+				((CVars::bxt_hud_jumpdistance.GetInt() == 1 ?
+					player.velocity[2] > 0.0f : player.velocity[2] != 0.0f) && prevVel[2] == 0.0f) ||
+				(player.velocity[2] > 0.0f && prevVel[2] < 0.0f)))
 			{
-				// 1 = jumpdistance will update when velocity is reasonably positive
-				// not 1 = jumpdistance will update when velocity changes, eg just falling
-				if (!inJump && (
-					((CVars::bxt_hud_jumpdistance.GetInt() == 1 ?
-						player.velocity[2] > 0.0f : player.velocity[2] != 0.0f) && prevVel[2] == 0.0f) ||
-					(player.velocity[2] > 0.0f && prevVel[2] < 0.0f)))
+				inJump = true;
+
+				// by the time the instantaneous velocity is here
+				// there is already displacement, which is in the previous frame
+				vecCopy(prevPlayerOrigin, prevOrigin);
+			}
+			else if (inJump && ((player.velocity[2] == 0.0f && prevVel[2] < 0.0f)))
+			{
+				inJump = false;
+
+				// add 16*2 because of player size can reach the edge of the block up to 16 unit
+				double distance = length(player.origin[0] - prevOrigin[0], player.origin[1] - prevOrigin[1]) + 32.0;
+
+				if ((distance > 150.0 && distance < 290.0) || CVars::bxt_hud_jumpdistance.GetInt() != 1) // from uq_jumpstats default
 				{
-					inJump = true;
-
-					// by the time the instantaneous velocity is here
-					// there is already displacement, which is in the previous frame
-					vecCopy(prevPlayerOrigin, prevOrigin);
+					jumpDistance = distance;
 				}
-				else if (inJump && ((player.velocity[2] == 0.0f && prevVel[2] < 0.0f)))
-				{
-					inJump = false;
-
-					// add 16*2 because of player size can reach the edge of the block up to 16 unit
-					double distance = length(player.origin[0] - prevOrigin[0], player.origin[1] - prevOrigin[1]) + 32.0;
-
-					if (distance > 150.0 || distance < 290.0) // from uq_jumpstats default
-					{
-						// +18f in case normal jump to duck
-						if (player.origin[2] == prevOrigin[2] || trunc(player.origin[2] + 18.0f - prevOrigin[2]) == 0.0f)
-						{
-							fadingFrom[0] = 0;
-							fadingFrom[1] = 255;
-							fadingFrom[2] = 0;
-						} else {
-							fadingFrom[0] = 255;
-							fadingFrom[1] = 0;
-							fadingFrom[2] = 0;
-						}
-
-						jumpDistance = distance;
-						passedTime = 0.0;
-					}
-				}
-				else if (player.origin[2] == prevPlayerOrigin[2])
-				{
-					// walking
-					inJump = false;
-				}
-
-				// Can be negative if we went back in time (for example, loaded a save).
-				double timeDelta = std::max(flTime - lastTime, 0.0f);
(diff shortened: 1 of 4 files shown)


khanghugo:
This won't be useful in game like HL where speed matters more but it is good for CS1.6 KZ TAS recording that might tell something's up ahead.

BunnymodXT/hud_custom.cpp
```diff
@@ -661,6 +661,95 @@ namespace CustomHud
...
+				}
+				else if (inJump && ((player.velocity[2] == 0.0f && prevVel[2] < 0.0f)))
+				{
+					inJump = false;
+
+					// add 16*2 because of player size can reach the edge of the block up to 16 unit
+					jumpDistance = length(player.origin[0] - prevOrigin[0], player.origin[1] - prevOrigin[1]) + 32.0;
+
+					// +18f in case normal jump to duck
+					if (player.origin[2] == prevOrigin[2] || (player.origin[2] + 18.0f) == prevOrigin[2])
```
YaLTeR, line 705: Does this exact equality actually work in practice, especially with the `+ 18.0`? Floats are pretty finicky...

BunnymodXT/hud_custom.cpp
```diff
@@ -661,6 +661,95 @@ namespace CustomHud
...
+					// by the time the instantaneous velocity is here
+					// there is already displacement, which is in the previous frame
+					vecCopy(prevPlayerOrigin, prevOrigin);
+				}
+				else if (inJump && ((player.velocity[2] == 0.0f && prevVel[2] < 0.0f)))
+				{
+					inJump = false;
+
+					// add 16*2 because of player size can reach the edge of the block up to 16 unit
+					jumpDistance = length(player.origin[0] - prevOrigin[0], player.origin[1] - prevOrigin[1]) + 32.0;
```
YaLTeR, line 702: This is kinda weird, like jumping in place will already give you 32 units. How do the kz stats plugins measure it?

BunnymodXT/hud_custom.cpp
```diff
@@ -661,6 +661,95 @@ namespace CustomHud
...
+					// +18f in case normal jump to duck
+					if (player.origin[2] == prevOrigin[2] || (player.origin[2] + 18.0f) == prevOrigin[2])
+					{
+						fadingFrom[0] = 0;
+						fadingFrom[1] = 255;
+						fadingFrom[2] = 0;
+					} else {
+						fadingFrom[0] = 255;
+						fadingFrom[1] = 0;
+						fadingFrom[2] = 0;
```
YaLTeR, line 713: I think it's weird that green/red color here will mean something different from the jump speed counter. Maybe we can use some other color here?

BunnymodXT/hud_custom.cpp
YaLTeR, line 705: This check will also fire if you standing-jump to a surface 18 units down.

khanghugo: It does work. It makes the number goes green when duck before ground. How do you want to do this instead?

BunnymodXT/hud_custom.cpp
khanghugo: https://kz-rush.com/en/article/calculating-distance

I did try using the value from `bxt_ch_get_pos` between the start and end of the jump and it reports the exact value of two places. It is what people might want to know how far they jump. But in practice, that would fall 32 units short, in comparison with how other kz plugins work. There is another addition of 32 to make it "block jump", rather than just a normal jump.

BunnymodXT/hud_custom.cpp
khanghugo: Green means that jump is a normal jump, start and end are on the same height. Red will means that the jump is not landing on the same height. If this makes it looks too weird then I guess we can just do no color.

BunnymodXT/hud_custom.cpp
khanghugo: That is true. I don't really know what to do for that case because it would require the info of the player position such as on ground or in duck and stuffs. This is definitely a step down but it is primitive, and the best solution I can think of right now without checking extras.

khanghugo: The number goes green means the jump is an actual jump on the same block height.

BunnymodXT/hud_custom.cpp
khanghugo: Oh and about jumping in place, I guess that should be an edge case. It definitely can get checked if that is needed.

khanghugo: Maybe it only updates when above certain distance, that would be better, I guess.

BunnymodXT/hud_custom.cpp
khanghugo: If we eliminate the color code thing then we don't need to compare float like above.

khanghugo:
Here is a small change that will check when to update the jump distance. I am still keeping the color thing because I am not sure how it should be proceeded. If there is no need for color, then all of the above concerns would be gone.

BunnymodXT/hud_custom.cpp
YaLTeR: If other kz plugins add the 32 then fine, although kinda weird because as that article points out, it's only correct for axis-aligned jumps and not for diagonal ones.

BunnymodXT/hud_custom.cpp
YaLTeR: Let's do no color for now

BunnymodXT/hud_custom.cpp
khanghugo: Yea that. Given what there is, calculating the distance of a diagonal jump would be really difficult too, at least without additional checking of other properties that `custom_hud` does not really need. But I from my limited mapping experience, making a significantly measurable diagonal jump block would be unnecessarily hard, especially doing the math. It is a possibility but I don't think any mapper will be willing to do that. Though, this is indeed interesting and I have to keep this fact in mind (as a suggestion for mappers to do uncanny jump).

BunnymodXT/hud_custom.cpp
khanghugo: Got that.

BunnymodXT/hud_custom.cpp
```diff
@@ -661,6 +661,58 @@ namespace CustomHud
...
+				vecCopy(prevPlayerOrigin, prevOrigin);
+			}
+			else if (inJump && ((player.velocity[2] == 0.0f && prevVel[2] < 0.0f)))
+			{
+				inJump = false;
+
+				// add 16*2 because of player size can reach the edge of the block up to 16 unit
+				double distance = length(player.origin[0] - prevOrigin[0], player.origin[1] - prevOrigin[1]) + 32.0;
+
+				if ((distance > 150.0 && distance < 290.0) || CVars::bxt_hud_jumpdistance.GetInt() != 1) // from uq_jumpstats default
```
YaLTeR, line 695: Is this 290 limit ok for hl1 tho? BXT is hl1-first after all

khanghugo: Whoops. I guess no limit seems okay due to recent incidents of water move stuffs. This is generally just to prevent speed randomly gets overwritten.

khanghugo:
So I think about axis align thing, maybe there could be a little bit of trigonometry could be applied to properly scale the 16 units of box properly, depeneding on the displacement vector of the jump. Let me try this and see

BunnymodXT/hud_custom.cpp
```diff
@@ -661,6 +661,58 @@ namespace CustomHud
...
+		if (CVars::bxt_hud_jumpdistance.GetBool())
+		{
+			static bool inJump = false;
+			static double jumpDistance = 0.0;
+
+			// 1 = jumpdistance will update when velocity is reasonably positive
+			// not 1 = jumpdistance will update when velocity changes, eg just falling
+			if (!inJump && (
+				((CVars::bxt_hud_jumpdistance.GetInt() == 1 ?
+					player.velocity[2] > 0.0f : player.velocity[2] != 0.0f) && prevVel[2] == 0.0f) ||
```
YaLTeR, line 679: Btw is the operator precedence right here? Pretty sure I got bitten by the `x ? y : z` precedence a few times with something similar to `a == b ? c : d` being parsed as `a == (b ? c : d)` or so... Could you add brackets around all of this to make sure?

YaLTeR:
Thanks!

PR #342, opened 2022-10-06
Reworked some autostops 

SmileyAG:
When I did `multi_manager` autostops 2 years ago for HL1 mods, I didn't thinked that in same modification there can be a few maps that uses a similiar name for entity, so I decided to download each fucking mod for which auto-stop is supported and researched in detail all `.bsp` files and got fixed some issues.

In CSCZDS if u using stop timer via firing `BLOOOM` in **multi_manager**, it can be easily triggered by player fall from level, so I redo with a more success idea that will not cause issues in future at all by hooking `CBreakable::Die`, which is informate about entity breaking with specific target name. (There is only 1 `func_breakable` entity with that specific **target**, so u don't need worries at all)

@chinese-soup Can you make sure that **CBreakable::Die** got hooked on Linux and auto-stop in CSCZDS works?

1) Get save: https://cdn.discordapp.com/attachments/842072844275154947/1027423776956227585/auto.sav
2) Start timer
3) Press `+attack` to explode tower
4) See if timer is stopped or not

### DIFF
--- a/BunnymodXT/discord_integration.cpp
+++ b/BunnymodXT/discord_integration.cpp
@@ -50,26 +50,6 @@ namespace discord_integration
 				}
 			}
 
-			inline static size_t get_map_name(char* dest, size_t count)
-			{
-				auto map_path = cl.pEngfuncs->pfnGetLevelName();
-
-				const char* slash = strrchr(map_path, '/');
-				if (!slash)
-					slash = map_path - 1;
-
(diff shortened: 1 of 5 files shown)


chinese-soup:
> @chinese-soup Can you make sure that CBreakable::Die got hooked on Linux and auto-stop in CSCZDS works?
lgtm

YaLTeR:
Thanks

PR #343, opened 2022-10-08
RPC: Added support to see difference between chapters on 'ba_teleport2' in Blue Shift

SmileyAG:


BunnymodXT/discord_integration.cpp
```diff
@@ -136,7 +137,7 @@ namespace discord_integration
 							// Adjust gameDir to lowercase
 							cl.ConvertToLowerCase(gd);
 
-							if (!strncmp(gd, "valve", 5) || !strncmp(gd, "abh", 3) || !strncmp(gd, "glitchless", 10))
+							if (!strncmp(gd, "valve", 5) || !strcmp(gd, "abh") || !strncmp(gd, "glitchless", 10))
```
YaLTeR, line 140: Why?

SmileyAG: I'm sure that some modifications may have a name for directory starting from `abh`, so better clamp it to **strcmp** tbh

YaLTeR:
Thanks

PR #344, opened 2022-10-11
Show 'MessageBox' if user downloaded a wrong version of BXT and null some pointers

SmileyAG:
`MessageBox` is a meaningless to use since it waits from user to press **OK** for close game, while you can just ignore window and still play it....

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -493,7 +493,7 @@ void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* modul
 		if (!is_cof) {
 			ClientDLL::GetInstance().pEngfuncs = nullptr;
 			ServerDLL::GetInstance().pEngfuncs = nullptr;
-			MessageBox(NULL, "Loaded BunnymodXT (CoF edition) to non-CoF game! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
+			MessageBox(NULL, "Loaded Bunnymod XT (CoF version) in non-CoF game! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
 		}
 		#else
 		if (is_cof) {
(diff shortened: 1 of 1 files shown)


YaLTeR:
Please make it a MessageBox without exiting. Exiting a process with no user-visible explanation does not help. And if the player decides to ignore a MessageBox and play anyway then that's their problem.

YaLTeR:
Sure, that sounds good.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -487,6 +487,22 @@ void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* modul
 			ORIG_CL_CheckGameDirectory, HOOKED_CL_CheckGameDirectory,
 			ORIG_SaveGameSlot, HOOKED_SaveGameSlot);
 	}
+
+	#ifdef _WIN32
+		#ifdef COF_BUILD
+		if (!is_cof) {
+			ClientDLL::GetInstance().pEngfuncs = nullptr;
+			ServerDLL::GetInstance().pEngfuncs = nullptr;
+			MessageBox(NULL, "Loaded BunnymodXT (CoF edition) to non-CoF game! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
```
YaLTeR, line 496:
```suggestion
			MessageBox(NULL, "Loaded Bunnymod XT (CoF version) in non-CoF game! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -487,6 +487,22 @@ void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* modul
...
+		}
+		#else
+		if (is_cof) {
+			ClientDLL::GetInstance().pEngfuncs = nullptr;
+			ServerDLL::GetInstance().pEngfuncs = nullptr;
+			MessageBox(NULL, "Loaded BunnymodXT (HL edition) to CoF game! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
```
YaLTeR, line 502:
```suggestion
			MessageBox(NULL, "Loaded BunnymodXT (HL version) in CoF! Download the right version!", "Fatal Error", MB_OK | MB_ICONERROR);
```

YaLTeR:
Thanks

PR #345, opened 2022-10-12
Added bxt_cof_slowdown_if_use_on_ground

SmileyAG:
**Cry of Fear** not have velocity restrict from `+use` on ground, so I added a BXT cvar that does same behaviour with `+use` as in HL games!


YaLTeR:
Thanks

PR #348, opened 2022-10-13
Some minor fix and additions to some HUDs

khanghugo:
This one
jumpspeed1

Old one
jumpspeed2


khanghugo:
Maybe my definition of jump speed is not agreeable. If so, this could be scrapped, or it can be `bxt_hud_jumpspeed 2`.

khanghugo:
Aww, just did something terrible.

khanghugo:
So this branch would cover two things because I accidentally force push some changes that I didn't really intend. I think both are reasonable enough.

PR #350, opened 2022-10-14
Add origins to bxt_hud_tas_editor_status

khanghugo:
I think these so far are what needed.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -77,6 +77,8 @@ namespace CustomHud
 
 		float vel;
 		float zvel;
+		float xpos;
+		float ypos;
 		float zpos;
 		float realyaw;
 		float health;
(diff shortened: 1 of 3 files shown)


khanghugo:
This one can wait a bit, I am adding more things and passing an array instead of whatever I am doing here...

BunnymodXT/hud_custom.cpp
```diff
@@ -544,9 +546,18 @@ namespace CustomHud
 			}
 			else
 			{
-				out << "X: " << player.origin[0] << "\n"
-					<< "Y: " << player.origin[1] << "\n"
-					<< "Z: " << player.origin[2];
+				if (frame_bulk_selected)
+				{
+					out << "X: " << frame_bulk_status.xpos << "\n"
```
YaLTeR, line 551: It's kinda weird how all frame bulk info is in the tas editor status hud, but just the origin is suddenly in the origin hud, and not always. I'd rather put the frame bulk origin into the tas editor status hud too.

khanghugo: Hmm, wouldn't that make the thing bit crowded? I am intending to have bxt_hud_velocity dumped into `frame_bulk_status` as well.

YaLTeR: Well, yeah, but it's better than this surprising behavior imo

khanghugo: Oh, never mind, I don't need the bxt_hud_velocity, forgot it is plainly there.

khanghugo: So this means this change to bxt_hud_origin should be reverted and have the frame bulk origin in bxt_hud_tas_editor_status instead?

YaLTeR: Yeah

YaLTeR:
Thanks

PR #351, opened 2022-10-14
Update hlstrafe (CS1.6 lgagst fix)

khanghugo:


YaLTeR:
I don't remember but it might need some change in HwDLL that sets the HLStrafe version. Or maybe it doesn't need any changes because it uses the max version constant. Could you double-check?

khanghugo:
Just browsed through HwDLL.

The HLStrafe version is set in the hlstrafe submodule. The HwDLL is pulling the constant from the said submodule instead if setting its own. I have BXT working fine and properly on my end when I change header of .hltas file (for their perspective version of course).

YaLTeR:
Yeah, I also double-checked the previous hlstrafe version update commit, seems like nothing else is needed.

PR #353, opened 2022-10-15
Restored 'give' command for CS 1.6

SmileyAG:

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -225,6 +225,8 @@ void ServerDLL::Clear()
 	ORIG_CTriggerVolume__Spawn_Linux = nullptr;
 	ORIG_CBasePlayer__ForceClientDllUpdate = nullptr;
 	ORIG_CBasePlayer__ForceClientDllUpdate_Linux = nullptr;
+	ORIG_CBasePlayer__GiveNamedItem = nullptr;
+	ORIG_CBasePlayer__GiveNamedItem_Linux = nullptr;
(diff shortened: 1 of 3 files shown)


khanghugo:
Very good addition. 

Starting a game with `deathmatch 1` would also initialize a listen server, which takes some extra time during playback and lots of time for second game simulator to catch up. This would also help with TAS optimizer multi-game simulation as everything is just instantly faster. The set up for a TAS would be minimally `sv_cheats 1` for `give`.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2220,37 +2237,39 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, CTriggerVolume__Spawn_Linux, void*, thisptr
 
 HOOK_DEF_1(ServerDLL, void, __cdecl, ClientCommand, edict_t*, pEntity)
 {
-#ifdef _WIN32
-	if (!ORIG_CBasePlayer__ForceClientDllUpdate) {
+	if (!pEngfuncs || !pEntity->pvPrivateData) {
```
YaLTeR, line 2240: Why does this check `pEntity->pvPrivateData` when it's never used below?

SmileyAG: Otherwise it would crash game and that check used in original `ClientCommand` as well: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/client.cpp#L507-L508

YaLTeR:
Thanks

PR #354, opened 2022-10-15
Added support for CoF 1.55 and some improvements for CoF Steam

### LINKED ISSUE #347: CoF 1.55: 'bxt_cof_disable_save_lock' not restoring monster origin
It's not a issue of BXT code, it's in game code itself, and I don't know how to fix that.

https://user-images.githubusercontent.com/58108407/195468756-f52d2ee1-3f63-41f2-9eae-3a7877fcca0a.mp4

SmileyAG:
Closes #347


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4012,9 +4027,7 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 	RegisterCVar(CVars::bxt_force_clear);
 	RegisterCVar(CVars::bxt_disable_gamedir_check_in_demo);
 	RegisterCVar(CVars::bxt_remove_fps_limit);
-
-	if (ORIG_Host_ValidSave && cofSaveHack)
-		RegisterCVar(CVars::bxt_cof_disable_save_lock);
+	RegisterCVar(CVars::bxt_cof_disable_save_lock);
```
YaLTeR, line 4030: I guess this should be `if (is_cof)`?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5311,10 +5324,17 @@ bool HwDLL::TryGettingAccurateInfo(float origin[3], float velocity[3], float& he
 	armorvalue = pl->v.armorvalue;
 	waterlevel = pl->v.waterlevel;
 
-	if (is_cof) {
+	if (is_cof || ServerDLL::GetInstance().is_cof_155 || ServerDLL::GetInstance().is_cof_10) {
```
YaLTeR, line 5327: Shouldn't `is_cof` include all of these? Otherwise it should be renamed to something else, like `is_cof_something`.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -298,12 +303,10 @@ bool ServerDLL::CanHook(const std::wstring& moduleFullName)
 
 	std::wstring folderName = GetFolderName(moduleFullName);
 
-	#ifdef COF_BUILD
 	// HACK: In Cry of Fear client and server dlls are in the same directory.
```
YaLTeR, line 306: Comment should be updated to note that it's only in some particular CoF versions that this is the case, and not in others.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2807,3 +2923,11 @@ HOOK_DEF_0(ServerDLL, void, __cdecl, PM_UnDuck)
 
 	ORIG_PM_UnDuck();
 }
+
+HOOK_DEF_1(ServerDLL, void, __cdecl, ShiftMonsters, Vector, origin)
+{
+	if (CVars::bxt_cof_disable_save_lock.GetBool())
```
YaLTeR, line 2929: Shouldn't this condition be "this is during a saveload" rather than "we unlocked saveloading functionality"?

BunnymodXT/modules/HwDLL.cpp
SmileyAG: No.

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: In all versions of `Cry of Fear` DLLs is placed in **cl_dlls** folder.

SmileyAG: > Why is the ifdef removed?

BXT (CoF edition) is used only for Steam version of that game, while a mod. versions is using a normal BXT as I explained before in DMs.

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: I don't think there is need to be added separate cvar for that feature tbh, so I just included it to `bxt_cof_disable_save_lock` cvar itself.

SmileyAG:
Removed 1.0 version since I don't see a practical useful for speedrunners, got only left support for a 1.55 since it can be installed on normal GoldSrc engine.

BunnymodXT/modules/ServerDLL.cpp
YaLTeR: What cvar? Didn't you say that resetting the monster position is intended functionality sometimes? With the current implementation, as far as I can tell, the reset will be prevented *always when bxt_cof_disable_save_lock is enabled* rather than *only on saveloads*.

YaLTeR: After discussing on Discord: looks like ShiftMonsters is only called on saveloads anyway.

YaLTeR:
Thanks

PR #356, opened 2022-10-16
Fix bxt_hud_jumpspeed accuracy with bhopcap

khanghugo:
My apologies. I hoped the other commit was fine until I tested this a bit and have seen some wrong things with it. 

When there is bhop cap, it will pull the speed from that jump frame, which is higher than the bhop cap wiill allow. It is inaccurate so this is the fix.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -603,7 +603,6 @@ namespace CustomHud
 			static double passedTime = FADE_DURATION_JUMPSPEED;
 			static int fadingFrom[3] = { hudColor[0], hudColor[1], hudColor[2] };
 			static double jumpSpeed = 0.0;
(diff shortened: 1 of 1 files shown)


BunnymodXT/hud_custom.cpp
```diff
@@ -603,6 +603,7 @@ namespace CustomHud
 			static double passedTime = FADE_DURATION_JUMPSPEED;
 			static int fadingFrom[3] = { hudColor[0], hudColor[1], hudColor[2] };
 			static double jumpSpeed = 0.0;
+			HLStrafe::MovementVars vars = HwDLL::GetInstance().GetMovementVars();
```
YaLTeR, line 606: Put this below, right before where it's used.

khanghugo:
Alright, this should be it. I forgot the `bxt_bhopcap` is a thing so that should be kilometers better.

YaLTeR:
Thanks

PR #333, opened 2022-09-18
Added experimental support for Cry of Fear

SmileyAG:

PR #357, opened 2022-10-16
Handle cases when player in water / on ladder for +bxt_tas_autojump

### LINKED ISSUE #339: Handle cases when player in water / on ladder for +bxt_tas_autojump

SmileyAG:
Closes #339 

All credits goes to hoboken, the reason why it requested from my account is because he asked me to do so.

Need to merge https://github.com/HLTAS/hlstrafe/pull/5 at first and I'll do rebase here to fix errors.

### DIFF
--- a/BunnymodXT/input_editor.cpp
+++ b/BunnymodXT/input_editor.cpp
@@ -64,7 +64,11 @@ void EditedInput::simulate() {
 		std::placeholders::_3
 	);
 
-	const auto point_contents_func = ClientDLL::GetInstance().pEngfuncs->PM_PointContents;
+	const auto point_contents_func = std::bind(
+		*(ClientDLL::GetInstance().pEngfuncs->PM_PointContents),
+		std::placeholders::_1,
+		nullptr
+	);
 
 	size_t total_frames = frame_bulk_starts[first_frame_bulk];
 
--- a/BunnymodXT/input_editor.cpp
+++ b/BunnymodXT/input_editor.cpp
@@ -64,6 +64,8 @@ void EditedInput::simulate() {
 		std::placeholders::_3
 	);
 
+	const auto point_contents_func = ClientDLL::GetInstance().pEngfuncs->PM_PointContents;
+
 	size_t total_frames = frame_bulk_starts[first_frame_bulk];
 
 	auto player = *(player_datas.cend() - 1);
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4175,6 +4175,8 @@ void HwDLL::InsertCommands()
 						player.InDuckAnimation = (pl->v.bInDuck != 0);
 						player.DuckTime = static_cast<float>(pl->v.flDuckTime);
 						player.StaminaTime = pl->v.fuser2;
+						player.WaterLevel = pl->v.waterlevel;
```
YaLTeR, line 4178: This value is 1 frame out of date. Instead of passing waterlevel from outside, it should be computed inside `GetPositionType` [the same way it's done in `CatagorizePosition`](https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/pm_shared/pm_shared.c#L1557), which is what `GetPositionType` is based on.

BunnymodXT/input_editor.cpp
```diff
@@ -64,6 +64,8 @@ void EditedInput::simulate() {
 		std::placeholders::_3
 	);
 
+	const auto point_contents_func = ClientDLL::GetInstance().pEngfuncs->PM_PointContents;
```
YaLTeR, line 67: Is `ClientDLL::GetInstance().pEngfuncs` set regardless of whether the client is found or not? Because thus far TASing functionality was independent of the Client DLL I think, and I'd like to keep it that way.

hobokenn: It is ["forwarded"](https://github.com/YaLTeR/BunnymodXT/blob/master/BunnymodXT/modules/HwDLL.cpp#L1319) from the engine, for lack of the better word, and set for every client.dll, the only time it wouldn't work is with an encrypted client. I thought it was fine since we already changed Get/Set Viewangles and had similar convo there https://github.com/YaLTeR/BunnymodXT/pull/333#discussion_r985154641

HLSDK/engine/cdll_int.h
```diff
@@ -190,7 +190,7 @@ typedef struct cl_enginefuncs_s
 	void						( *V_CalcShake )			( void );
 	void						( *V_ApplyShake )			( float *origin, float *angles, float factor );
 
-	int							( *PM_PointContents )		( float *point, int *truecontents );
+	int							( *PM_PointContents )		( const float *point, int *truecontents );
```
YaLTeR, line 193: Did you check in ghidra that it doesn't modify `point`? Just in case

hobokenn: yeah, it does not

PR #359, opened 2022-10-17
Fixed case when bxt_remove_fps_limit code is active if cheats disabled

SmileyAG:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5590,15 +5590,13 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
-			if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75)
-				is_0x75 = true;
+		if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75)
+			is_0x75 = true;
 
-			if (((*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x7E) && CVars::sv_cheats.GetBool()) || (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75))
-				MemUtils::ReplaceBytes(reinterpret_cast<void*>(pHost_FilterTime_FPS_Cap_Byte), 1, reinterpret_cast<const byte*>("\xEB"));
-		}
-		else if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0xEB)
+		if ((((*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x7E) && CVars::sv_cheats.GetBool()) || (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75)) && CVars::bxt_remove_fps_limit.GetBool())
```
YaLTeR, line 5596: Can you replace all huge `*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) ==` in conditions with variables like `is_0x75`? Then it'll be readable enough to figure out how to simplify it, because this is not reasonable.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5586,24 +5586,31 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
 	if (pHost_FilterTime_FPS_Cap_Byte)
 	{
-		if (CVars::bxt_remove_fps_limit.GetBool())
-		{
-			if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75)
-				is_0x75 = true;
-
-			if (((*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x7E) && CVars::sv_cheats.GetBool()) || (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75))
-				MemUtils::ReplaceBytes(reinterpret_cast<void*>(pHost_FilterTime_FPS_Cap_Byte), 1, reinterpret_cast<const byte*>("\xEB"));
+		if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x7E) {
```
YaLTeR, line 5596: You can do `const auto x = *reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte);` and then in the checks do `if (x == 0x7E)` etc instead of writing it out multiple times.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5586,24 +5586,31 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
+			is_0x7E = true;
+			is_0x7E_on_init = true;
+		} else if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75) {
+			is_0x75 = true;
+		} else if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0xEB) {
+			is_0xEB = true;
 		}
-		else if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0xEB)
+
+		if (CVars::bxt_remove_fps_limit.GetBool() && ((is_0x7E && CVars::sv_cheats.GetBool()) || is_0x75))
```
YaLTeR, line 5605: Why is the sv_cheats check only in 0x7E case?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5586,24 +5586,31 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
+			MemUtils::ReplaceBytes(reinterpret_cast<void*>(pHost_FilterTime_FPS_Cap_Byte), 1, reinterpret_cast<const byte*>("\xEB"));
+
+		if (is_0xEB && ((is_0x7E_on_init && !CVars::sv_cheats.GetBool()) || !CVars::bxt_remove_fps_limit.GetBool()))
```
YaLTeR, line 5608: I would restructure this as
```
if (bxt_remove_fps_limit && sv_cheats) {
  enable the thing
} else {
  disable the thing
}
```

YaLTeR (changes requested):

BunnymodXT/modules/HwDLL.cpp
SmileyAG: Uh, let me explain what these two cases do:

- *0x75* (1712-WON pattern) unlocks a cap for a higher FPS (not needed in `sv_cheats` tbh)
- *0x7E* (Steam pattern) unlocks a cap for set a <20 FPS if `maxplayers > 1` (it really needs `sv_cheats` as I don't want anyone to use it on servers, the reason why I even added this is for local sessions with friends in co-op modifications for GoldSrc engine, e.g.: https://www.moddb.com/mods/blue-shift-coop/downloads)

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5586,24 +5586,22 @@ HOOK_DEF_1(HwDLL, int, __cdecl, Host_FilterTime, float, passedTime)
...
+	static bool is_0x7E_on_init = false;
 	if (pHost_FilterTime_FPS_Cap_Byte)
 	{
-		if (CVars::bxt_remove_fps_limit.GetBool())
-		{
-			if (*reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte) == 0x75)
-				is_0x75 = true;
+		const auto pByte = *reinterpret_cast<byte*>(pHost_FilterTime_FPS_Cap_Byte);
+		if (pByte == 0x7E)
+			is_0x7E_on_init = true;
```
YaLTeR, line 5594:
```suggestion
		static bool is_0x7E_on_init = (pByte == 0x7E);
```
and remove the declaration above

YaLTeR:
Alright, I guess it's good now. Thanks

PR #360, opened 2022-10-18
Added Bunnysplit MQ for Linux

chinese-soup:


BunnymodXT/Linux/interprocess.cpp
```diff
@@ -62,13 +63,24 @@ namespace Interprocess
...
-
 		std::vector<char> buf(10);
 		buf[0] = static_cast<char>(buf.size());
 		buf[1] = static_cast<char>(MessageType::TIME);
 		AddTimeToBuffer(buf.data() + 2, time);
+		
+		if (CVars::_bxt_bunnysplit_time_update_frequency.GetFloat() > 0.0f) {
+			static auto last_time = std::chrono::steady_clock::now() - std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()) + 1);
+			auto now = std::chrono::steady_clock::now();
+			if (now >= last_time + std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()))) {
```
YaLTeR, line 74: Can you extract variable on the `std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()))`?

YaLTeR (commented):
What happens if there's no client to read from the message queue? The tasview one will block once it exhausts the buffer, but what happens with the bunnysplit one? Does it start erroring and recreating the message queue every few seconds? If so, doesn't that cause lag? Most of the time there's no bunnysplit on linux so that's why I'm wondering, that situation shouldn't regress.

BunnymodXT/Linux/interprocess.cpp
chinese-soup: > What happens if there's no client to read from the message queue? The tasview one will block once it exhausts the buffer, but what happens with the bunnysplit one? Does it start erroring and recreating the message queue every few seconds? If so, doesn't that cause lag? Most of the time there's no bunnysplit on linux so that's why I'm wondering, that situation shouldn't regress.

Uhh, isn't this already answered in the OP? And no, it doesn't restart the message queue every few seconds.

chinese-soup: > Can you extract variable on the `std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()))`?

Not sure which do you mean by that, this one?
```
auto blah = CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()
static auto last_time = std::chrono::steady_clock::now() - std::chrono::milliseconds(static_cast<long long>(1000 / blah + 1);
// [...]
if (now >= last_time + std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()))) {
```

If so, should I backport it to the Windows version too?

chinese-soup: > Can you extract variable on the `std::chrono::milliseconds(static_cast<long long>(1000 / CVars::_bxt_bunnysplit_time_update_frequency.GetFloat()))`?

Done. Couldn't extract the whole `std::chrono::milisecond` call without doubling code, because then I wouldn't be able to add + 1 to it in `static auto last_time = //...`

Should I backport it to the Windows version too?

YaLTeR: > Uhh, isn't this already answered in the OP? And no, it doesn't restart the message queue every few seconds.

Well I still don't understand what happens when there's no reader then? Do the messages get dropped or what happens?

> Couldn't extract the whole std::chrono::milisecond call without doubling code, because then I wouldn't be able to add + 1 to it in static auto last_time = //...

Now that you extracted it I can actually see there's a + 1 there that I missed before, heh.

> Should I backport it to the Windows version too?

Sure.

chinese-soup: > Now that you extracted it I can actually see there's a + 1 there that I missed before, heh.
Same tbh, so optimized for humans now :+1: 

> Well I still don't understand what happens when there's no reader then? Do the messages get dropped or what happens?

Ah right, well, you are correct in the fact that currently it keeps on retrying to do mq_send and closed and opens the message queue, every time that `WriteBunnySplit` is called (aka most of the time when _bxt_bunnysplit_time_update_frequency is hit in `WriteTime`)

How would do you want me to solve this? Should it just fail (close the queue) and not reopen it automatically and only reopen it when `_bxt_interprocess_reset` is manually entered in the game console? 

> Sure.
Alright, will do.

YaLTeR: > How would do you want me to solve this? Should it just fail (close the queue) and not reopen it automatically and only reopen it when _bxt_interprocess_reset is manually entered in the game console?

Yeah idk this stuff is complicated, let's go with the "only reopen it when _bxt_interprocess_reset" solution for now

PR #362, opened 2022-10-28
Fix bxt_hud_jumspeed saga

khanghugo:
I think this is the end to it. Works great.

### DIFF
--- a/BunnymodXT/hud_custom.cpp
+++ b/BunnymodXT/hud_custom.cpp
@@ -614,7 +614,7 @@ namespace CustomHud
 					|| (player.velocity[2] > 0.0f && prevVel[2] < 0.0f))
 				{
 					double speed = length(prevVel[0], prevVel[1]);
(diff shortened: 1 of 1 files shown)


BunnymodXT/hud_custom.cpp
```diff
@@ -603,18 +603,20 @@ namespace CustomHud
...
+			bool ducktap = false;
 
 			int r = hudColor[0],
 				g = hudColor[1],
 				b = hudColor[2];
 
 			if (FADE_DURATION_JUMPSPEED > 0.0f)
 			{
-				if ((player.velocity[2] != 0.0f && prevVel[2] == 0.0f)
+				if ((ducktap = (player.velocity[2] < 0.0f && prevVel[2] == 0.0f))
```
YaLTeR, line 614: uhh, no assignments in conditionals please

khanghugo: I love sepples.

YaLTeR:
Thanks

PR #367, opened 2022-11-08
Add bxt_tas_editor_set_frametime

khanghugo:


BunnymodXT/hud_custom.cpp
```diff
@@ -1342,6 +1344,14 @@ namespace CustomHud
 
 			out << "Health: " << frame_bulk_status.health << '\n';
 			out << "Armor: " << frame_bulk_status.armor << '\n';
+
+			if (!(!frame_bulk_status.frametime.compare("0.010000001") ||
+					!frame_bulk_status.frametime.compare("0.001") || 
+					!frame_bulk_status.frametime.compare("0.002") || 
+					!frame_bulk_status.frametime.compare("0.004")
+					)) {
```
YaLTeR, line 1352: Why the condition? I think just print it maybe

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3578,6 +3580,23 @@ struct HwDLL::Cmd_BXT_TAS_Editor_Toggle
...
 
+struct HwDLL::Cmd_BXT_TAS_Editor_Set_Frametime
+{
+	USAGE("Usage: bxt_tas_editor_set_frametime <frametime>\n Sets frametime on the currently selected point.\n");
+
+	static void handler(const char *value)
+	{
+		auto& hw = HwDLL::GetInstance();
+
+		if (boost::lexical_cast<double>(value) <= 0.f) {
```
YaLTeR, line 3591: Can we avoid including `lexical_cast` just for this? Maybe `atof`, or honestly just `!isdigit(value[0])`

BunnymodXT/hud_custom.cpp
khanghugo: Hmm I thought to not add anything more to the TAS editor status as there is too much. Sure, I will just have it there.

BunnymodXT/modules/HwDLL.cpp
khanghugo: Thanks. I don't know how float works and concerned that 0.01000001 thing might get lost. Not anymore.

YaLTeR:
Thanks

PR #371, opened 2022-11-11
bxt_ch_get_pos also prints bxt_cam_fixed template

khanghugo:
Good for when as to decide even make thirdperson view.


YaLTeR:
Thanks

PR #373, opened 2022-11-17
Reworked a functional of some cosmetic cvars, added bxt_shake_remove

SmileyAG:
gordon

Notes:

- `bxt_shake_remove` added in the reason that there is still some ppls who not aware of -rs version
- Changed order of `MemUtils::MarkAsExecutable` calls to make it in same order as in **HwDLL.hpp**


YaLTeR:
Thanks

PR #376, opened 2022-12-02
Updated README and GitHub Actions

SmileyAG:


README.md
```diff
@@ -40,7 +40,7 @@ Use the [Bunnymod XT Launcher](https://github.com/YaLTeR/bxt-launcher).
 
 Building on Windows requires
 
-- Visual Studio 2017 or 2019
+- Visual Studio 2017, 2019 or 2022
```
YaLTeR, line 43:
```suggestion
- Visual Studio 2019 or 2022
```
The CI builds on 2019 so we can't promise it builds on 2017.

BunnymodXT/discord_integration.cpp
```diff
@@ -97,19 +97,15 @@ namespace discord_integration
 						const char* gameDir = cl.pEngfuncs->pfnGetGameDirectory();
 						// Get the map name and icon.
 						cl.GetMapName(map_name, ARRAYSIZE_HL(map_name));
-						if (gameDir && gameDir[0] && map_name[0])
+						if (gameDir && gameDir[0] && map_name[0] && hw.ORIG_build_number)
```
YaLTeR, line 100: Why? Could you please write explanations for these types of changes?

YaLTeR (changes requested):
For the CI, you can also replace the Rust actions like this:
```
      - name: Install Rust
        run: |
          rustup toolchain install stable --profile minimal
          rustup override set stable
          rustup target add i686-unknown-linux-gnu
```

BunnymodXT/discord_integration.cpp
SmileyAG: That's already described in name of commit: `Don't update RPC if 'build_number' function not found`

Because I don't want to update RPC if pattern for output number of engine build is missed.

YaLTeR: Yes, I do read commit messages. I don't understand why you don't want to update in this case. There's more info there than just the build number.

SmileyAG: > Yes, I do read commit messages. I don't understand why you don't want to update in this case. There's more info there than just the build number.

For all of that more information checks is already here, like `cl.pEngfuncs` and `gamedir[0] && map_name[0]`

The only thing was missed is checking for found pattern of `build_number`, because I want to know some of that specific info like.

YaLTeR: Yes, so why disable it altogether if there's plenty of info we can show without build_number?

SmileyAG: > Yes, so why disable it altogether if there's plenty of info we can show without build_number?

I rather preferring to have that check at that line, since if that not success to update RPC, then user would told version of engine that is unsupported, so we can add proper support to it later.

YaLTeR: Huh, why? What does showing map and game in discord rpc have to do with indicating to the user they are running an "unsupported build", whatever that means (I have no idea what you mean by this)?

YaLTeR:
Thanks!

PR #377, opened 2022-12-02
Refactor of cvar defining 

SmileyAG:


BunnymodXT/cvars.cpp
```diff
@@ -1,389 +1,13 @@
 #include "stdafx.hpp"
-
```
YaLTeR, line 2: unrelated change

BunnymodXT/cvars.hpp
```diff
@@ -1,5 +1,202 @@
...
+	X(bxt_viewmodel_ofs_up, "0") \
+	X(bxt_viewmodel_bob_angled, "0") \
+	X(bxt_show_bullets, "0") \
+	X(bxt_show_bullets_enemy, "0") \
+	X(bxt_anglespeed_cap, "1") \
+	X(bxt_speed_scaling, "1")
+
+#define FIND_WRAPPER(name) CVarWrapper name;
+#define DEFINE_WRAPPER(name, string) CVarWrapper name(#name, string);
+#define ADD_TO_FIND_LIST(name) &name,
```
YaLTeR, line 195: I'd like these "argument" macros to be defined right before they are used for clarity.

YaLTeR (changes requested):

BunnymodXT/cvars.cpp
SmileyAG: I don't think so. It looks more clean when a few `#include` goes in order without a some empty line.

YaLTeR: 1. stdafx is special because it is a precompiled header. It must always be the first include and it must be without any path even though it's not literally in the same folder. It's separated by a blank line to indicate this special nature.
2. The PR is about reworking cvar definition and the commit is called "Rework of cvar defining", nowhere does it mention include aesthetics.

YaLTeR:
Thanks!

PR #380, opened 2022-12-15
Visual changes

### LINKED ISSUE #292: Refactor bxt_show_triggers_legacy to make it toggable for demos

SmileyAG:
Closes #292

I know that I removed the alpha pulse mode for legacy triggers, but I don't want to think about a workaround to make it work correctly in `HUD_AddEntity`

That changes is a worth, because for example I can make a segment with that commands (TriAPI triggers have a code to pulsing them)

- bxt_show_triggers 1
- bxt_show_triggers_legacy 1
- bxt_show_triggers_legacy_alpha 0

And then I draw legacy triggers in demo with changing of `bxt_show_triggers_legacy_alpha` command, quite simple.


SmileyAG:
I decided to change the concept of that request a bit, and I added a lot of commands for toggling each visual element (e.g. that can be disabled: world, particles, skybox, water, brush-model entities, studio-model entities, sprite-model entities, players and their corpses)

All of that changes makes sense, for example it can give a huge increase of FPS if you at least disable draw of studio entities or water

Or disabling some of that visual stuff can just give good visibility, which makes it easier to segmenting/routing/TASing runs

Using `r_drawentities` not suit me at all, cuz it disables entities of all model types, while I want to make a sorting instead

Using `r_drawviewmodel` (instead of `bxt_remove_viewmodel`) is also not suitable well, because as we know, some modifications (example: CS 1.6) like to force the value of this command

The only reason for adding changes related to draw/hide players and their corpses is because it can be used to render demos in multiplayer modifications with closed source code

Don't mind that there checks of **ent->player** for hiding player models if `bxt_disable_world` or `bxt_disable_brush_entities` enabled, I made in case if some evil would inject BXT to multiplayer sessions so he wouldn't have a chance to get wallhack with those commands.

SmileyAG:
Got added draw of `func_ladder` entity to **bxt_show_triggers** commands, since as technically it based on `CBaseTrigger` class in the game code (https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/triggers.cpp#L1743)

BunnymodXT/cvars.hpp
```diff
@@ -55,6 +55,9 @@
 	X(bxt_disable_gamedir_check_in_demo, "0") \
 	X(bxt_remove_fps_limit, "0") \
 	X(bxt_cof_disable_save_lock, "0") \
+	X(bxt_remove_viewmodel, "0") \
+	X(bxt_disable_world, "0") \
+	X(bxt_disable_particles, "0") \
```
YaLTeR, line 60: You seem to have intended to split this PR into well-separated commits, however this change for example is unrelated to this commit and should be in a different commit. Are you going to fix this?

SmileyAG: Yeah that's was a mistake while rebasing to latest master to add all cvars in that first commit, but I don't want to mess around with commits for reverting changes so I'm not going to fix it.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2156,64 +2156,73 @@ void ServerDLL::DoAutoStopTasks()
...
 		r = 203;
 		g = 103;
 		b = 212;
-	} else {
+	} else if (std::strncmp(classname, "trigger_", 8) == 0) {
 		// White
 		r = 255;
 		g = 255;
 		b = 255;
+	} else if (std::strcmp(classname, "func_ladder") == 0) {
```
YaLTeR, line 2204: You can leave most of this function unchanged if you add an early is_ladder => return color.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2297,8 +2307,9 @@ HOOK_DEF_7(ServerDLL, int, __cdecl, AddToFullPack, struct entity_state_s*, state
 				ent->v.rendermode = kRenderTransTexture;
 		}
 	}
-	else if (is_trigger && CVars::bxt_show_triggers_legacy.GetBool()) {
-		ent->v.effects &= ~EF_NODRAW;
+	else if ((is_trigger || is_ladder) && CVars::bxt_show_triggers_legacy.GetBool()) {
+		if (is_trigger)
```
YaLTeR, line 2311: Why is this check needed?

BunnymodXT/modules/ServerDLL.cpp
khanghugo: There is a CC warnings from this.

YaLTeR:
Thanks

PR #382, opened 2022-12-17
Add target_yaw look_at and goodies

khanghugo:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3628,6 +3628,42 @@ struct HwDLL::Cmd_BXT_TAS_Editor_Set_Change_Type
...
+{
+	USAGE("Usage: bxt_tas_editor_set_target_yaw_type <type>\n Set type of target_yaw for a point in the camera editor. Valid types (currently supported) are velocity_lock, look_at, and velocity.\n");
+
+	static void handler(const char *what)
+	{
+		auto& hw = HwDLL::GetInstance();
+
+		if (!strcmp(what, "velocity_lock")) {
+			hw.tas_editor_set_target_yaw_velocity_lock = true;
+		} else if (!strcmp(what, "velocity")) {
```
YaLTeR, line 3641: Is this needed?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3628,6 +3628,42 @@ struct HwDLL::Cmd_BXT_TAS_Editor_Set_Change_Type
...
+				hw.tas_editor_set_target_yaw_look_at = true;
+				hw.tas_editor_set_target_yaw_look_at_entity = entity;
+
+				if (scan_entity == 4) {
+					hw.tas_editor_set_target_yaw_look_at_x = x;
+					hw.tas_editor_set_target_yaw_look_at_y = y;
+					hw.tas_editor_set_target_yaw_look_at_z = z;
+				}
+			} else if (sscanf(what, "look_at %f %f %f", &x, &y, &z) == 3) {
+				hw.tas_editor_set_target_yaw_look_at = true;
```
YaLTeR, line 3658: Shouldn't this set entity to 0?

YaLTeR (changes requested):

BunnymodXT/modules/HwDLL.cpp
khanghugo: I think the cvar is needed so there would be no need to reload the script. Is target_yaw velocity needed? No. I just put it there because things come in three.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4320,6 +4356,27 @@ void HwDLL::InsertCommands()
...
+
+						StrafeState.TargetYawLookAtOrigin[0] = origin[0];
+						StrafeState.TargetYawLookAtOrigin[1] = origin[1];
+						StrafeState.TargetYawLookAtOrigin[2] = origin[2];
+					}
+				} else {
+					StrafeState.TargetYawLookAtOrigin[0] = 0;
+					StrafeState.TargetYawLookAtOrigin[1] = 0;
+					StrafeState.TargetYawLookAtOrigin[2] = 0;
+				}
```
YaLTeR, line 4378: Put this unconditionally above the if. This way you avoid bugs, like the one already present in this code where you don't set to 0 when entity is > num edicts

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3628,6 +3628,41 @@ struct HwDLL::Cmd_BXT_TAS_Editor_Set_Change_Type
 	}
 };
 
+struct HwDLL::Cmd_BXT_TAS_Editor_Set_Target_Yaw_Type
+{
+	USAGE("Usage: bxt_tas_editor_set_target_yaw_type <type>\n Set type of target_yaw for a point in the camera editor. Valid types (currently supported) are velocity_lock, look_at, and velocity.\n");
```
YaLTeR, line 3633:
```suggestion
	USAGE("Usage: bxt_tas_editor_set_target_yaw_type <type>\n Set type of target_yaw for a point in the camera editor. Valid types (currently supported) are velocity_lock and look_at [entity <index>] [<x> <y> <z>].\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3628,6 +3628,41 @@ struct HwDLL::Cmd_BXT_TAS_Editor_Set_Change_Type
...
+		} else {
+			unsigned int entity;
+			float x, y, z;
+
+			int scan_entity = sscanf(what, "look_at entity %d %f %f %f", &entity, &x, &y, &z);
+			if (scan_entity) {
+				hw.tas_editor_set_target_yaw_look_at = true;
+				hw.tas_editor_set_target_yaw_look_at_entity = entity;
+
+				if (scan_entity == 4) {
```
YaLTeR, line 3650: Set the coordinates to 0 unconditionally before this if.

BunnymodXT/triangle_drawing.cpp
```diff
@@ -723,6 +723,7 @@ namespace TriangleDrawing
 							break;
 						case HLTAS::ConstraintsType::YAW:
 						case HLTAS::ConstraintsType::YAW_RANGE:
+						case HLTAS::ConstraintsType::LOOK_AT:
```
YaLTeR, line 726: Color it something different I think

YaLTeR (changes requested):

BunnymodXT/modules/HwDLL.cpp
khanghugo: Is it too much for me to believe that C has sane default?

YaLTeR: I mean float x = 0, y = 0, z = 0;

YaLTeR:
Thanks!

PR #386, opened 2023-01-03
Add bxt_tas_editor_force_mark_as_stale

khanghugo:
Especially useful when some weird RNG stuffs happen somewhere before the current framebulk and it is just movement section and it is in EDIT mode and the framebulk is nowhere to be seen.

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -3495,9 +3495,9 @@ struct HwDLL::Cmd_Minus_BXT_TAS_Editor_Insert_Point
 	}
 };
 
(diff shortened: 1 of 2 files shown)


YaLTeR:
A better name would be something like bxt_tas_editor_resimulate.

Do I understand correctly this is useful only when using multiple games?

khanghugo:
Yea, only useful when simulating from another game. While I was making an RNG dependent TAS, sometimes later changes just desyncs the whole thing (while I just want to edit a few small changes). The desync sometimes happens when loading the script too (even no RNG involved).

Sure, I will change the name.

YaLTeR:
Thanks

PR #387, opened 2023-01-06
EntityInfo: draw damage for func_door/func_rotating/func_train

SmileyAG:
Looking to HLSDK code, I realized that not only doors can give a infinite health for blocked player if `dmg` set to negative value, `func_rotating` and `func_train` entities can do that as well.


YaLTeR:
Thanks

PR #389, opened 2023-01-06
StudioModelRenderer: Don't use same offsets for all mods 

SmileyAG:


BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -419,51 +420,38 @@ void ClientDLL::FindStuff()
 			}
 		});
 
-	auto fStudioCalcAttachments = FindAsync(
-		ORIG_StudioCalcAttachments,
-		patterns::client::StudioCalcAttachments,
+	auto fCStudioModelRenderer__StudioCalcAttachments = FindAsync(
```
YaLTeR, line 459: Same as the other PR -- you seem to have intended to split the PR into commits, however these changes certainly don't fall under "Renamed 'StudioCalcAttachments' function". Do you intend to fix this?

SmileyAG: That changes I certainly assumed to put into `Renamed 'StudioCalcAttachments' function` commit, so I don't see wrong here.

YaLTeR: I meant as in the things below which introduce offVectorTransform and change the offset calculation. I guess otherwise with the new name it would get too unwieldy? In those cases what you'd do is first make a commit with the offVectorTransform change, then make a commit titled "Rename X" which only renames X and does not otherwise change the code flow.

SmileyAG: I intentionally decided to push that changes into one commit instead of splitting, but yeah I picked not good name for that commit, sorry.

YaLTeR:
Thanks

PR #393, opened 2023-01-08
CoF 1.0-1.2: Hotfix offsets for ForceClientDllUpdate

SmileyAG:
In generally it's not my fault that in the previous pull request I submitted incorrect offsets for CoF 1.0-1.2 versions, this is all due to awful implementation of .pdb support in `Ghidra`

I always look for offsets manually in functions where it uses and for all other mods I double-checked them many times and can vouch for their correctness 

But for CoF, developers attached a .pdb file for each patch version of the game, so I decided to rely on the offset-variable list that `Ghidra` was able to pull it out of the .pdb file and I realized that she does it terribly

In `IDA Pro` there is good support of .pdb file since it does a proper work to point offsets to variables in class struct, while Ghidra really fucking sucks in it so I can't trust it anymore for DLLs with .pdb file

TLDR: Fixed wrong offsets for CoF 1.0-1.2 that it made only due of awful support of .pdb file in Ghidra, don't worry about offsets in other HL modifications/versions cuz they are correct and I double-checked them.


YaLTeR:
Thanks, good to know

PR #401, opened 2023-01-14
Add patterns for AG 6.7

fireblizzard:


SmileyAG:
May I ask from what commit you compiled AG 6.7 and what build tools version used? 

I used compiled AG 6.7 build from GitHub Actions and patterns are different, commit: https://github.com/fireblizzard/agmod/commit/d6d2672e6ffd31d361b5d0245f7536ffae5d6bdf (Linker version: 14.29, `v142` build tools)

SmileyAG:
Oh btw, isn't it would be simple in AG 6.7 code to make a exported functions that needed for BXT? So we can avoid to make a extra patterns for it.

fireblizzard:
> May I ask from what commit you compiled AG 6.7 and what build tools version used?
> 
> I used compiled AG 6.7 build from GitHub Actions and patterns are different, commit: [fireblizzard/agmod@d6d2672](https://github.com/fireblizzard/agmod/commit/d6d2672e6ffd31d361b5d0245f7536ffae5d6bdf) (linker version: 14.29, `v142` build tools)
> 
> GitHub Actions build: https://github.com/fireblizzard/agmod/actions/runs/3814621849

It was a local build, I didn't even remember we have automatic builds on that repo.
It's built with the previous commit (doesn't matter since the last commit is just text), same toolset (v142, 14.29.30133)

fireblizzard:
I think we're gonna continue with my local build for AG speedrunning, unless you wanna commit the required changes for using the symbols instead of patterns. Ghidra is not able to find a few functions on the GH actions build even with the PDB (all code pointing to those is also unreachable) and i wouldn't like to delay the next release further + i dont have a lot of time left today for this

YaLTeR:
Why are you using BXT with AG btw?

> Oh btw, isn't it would be easier to make in AG 6.7 a functions that needed for BXT as exported? So we can avoid to make a a lot of extra patterns and hook it by symbols instead

Yeah I think that would make sense.

> I'm also think about to make pull request in some other HLSDK codebases as halflife-updated that been using rn by game modders for make some functions as exported, so we don't need to waste time with finding patterns or extend them for a new released mods.

Just saying, in my opinion it's not really 100% warranted; if I was a maintainer of one of these I'd want a good justification. Like, for something like time tracking I would agree, but if it's exporting something to change the HUD color then I probably wouldn't, idk

fireblizzard:
> Why are you using BXT with AG btw?

Mostly for meme runs on twitch, and for people who are curious about how the HL campaign plays with AG rules.

To add more context: the AG speedrunning package has always used my locally compiled builds because we did not have GH actions set up, but i should really use that build now that we have it, just in case anyone has doubts about the dll. I also gave them a custom BXT with AG patterns so that they can use most of the commands. If the patterns were in the official BXT build, then we could consider adding a category for these AG runs or making AG's own src com page be used for these runs. I guess we can also do that with the custom BXT build but it would be bad to have several versions of BXT around. So my plan is to export those functions in AG at some point after this release and then support those symbols in BXT, and i don't really have time to test that if you wanna release the next version ASAP, that's why i would like to go with the current state of the PR. I can understand that it's a bit silly merging this PR if these patterns are gonna be useless within 6 months or whatever, but i mean i dont think it's gonna hurt either.

YaLTeR:
> the AG speedrunning package has always used my locally compiled builds because we did not have GH actions set up, but i should really use that build now that we have it, just in case anyone has doubts about the dll

Yeah, it's also helpful for avoiding mistakes, like uncommitted changes in your local tree when you're building the release build, or whatever.

As for this PR, tbh I think y'all can just use a custom BXT build for now? The release is mostly important because we declare only released build can be used for submitting runs to the official boards, but AG isn't there yet. So there's no rush to implement it in a better way?

fireblizzard:
My idea was to submit the category to SRC after doing an "official" release of AG 6.7, which may happen in the next couple months after i make up my mind about a couple things. But then it may have some code change that breaks patterns... so yeah let's leave this PR for another release, when i have my AG stuff sorted out and i can implement BXT support for it properly through symbols :)

SmileyAG:
Sorry for the message after a long time, but I want to say that pull request is closed now in favor of redo it by exporting functions on SDK side to make it used in a future BXT codebase (there will would be a documentation on how to do it)

Because having too many hardcoded patterns and offsets is a really-really-really bad idea, even though you are ready to do it yourself and not ask anyone for help, it’s still better to leave that choice for manual support only for modification/versions that commonly used and also doesn't have the public source code, otherwise make your mod adapted to the future BXT codebase instead that would read that data if it found

PR #405, opened 2023-01-16
Added bxt_cof_disable_viewpunch_from_jump

SmileyAG:
I know that `bxt_remove_punchangles` exist, but I want to remove it for such a special case and not for all others, cuz it annoys me a that the camera punches when jumping.


YaLTeR:
Thanks

PR #409, opened 2023-01-23
RPC: add different map names for older versions of CoF

SmileyAG:
In first versions of `Cry of Fear` maps starting with prefix: `cof_campaign_01_p` have a different names instead.


YaLTeR:
Thanks

PR #411, opened 2023-01-23
Moved 'InitGraph' under the 'ForceClientDllUpdate' to make bxt_show_nodes draw a right nodes in CSCZDS

SmileyAG:
*bruuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuh*

Well, seems to be that InitGraph before was called earlier than ForceClientDllUpdate, and some variables due of that as `is_czeror` not being set and check with it not using due of that


YaLTeR:
Thanks

PR #413, opened 2023-01-31
hud_custom: show only server time if 'bxt_hud_quickgauss' is 2

SmileyAG:


YaLTeR:
Why?

SmileyAG:
> Why?

Looking to server time is good for some cases when I'm doing some of routing or tests in other mods, and I don't want to draw other stats that related to Quickgauss glitch

YaLTeR:
Alright

PR #414, opened 2023-01-31
Added bxt_collision_depth_map_fov

SmileyAG:


YaLTeR:
Hm, why would you want a separate cvar for this? Just change default_fov?

SmileyAG:
> Hm, why would you want a separate cvar for this? Just change default_fov?

- In some modifications `default_fov` is not exist

- Even if I use `bxt_force_fov` with hooked bxt-rs, then code with getting value from `bxt_force_fov` in BXT code would be ignored for some reason

YaLTeR:
> w/o hooked bxt-rs it works, probably bxt-rs just re-register that cvar again due it have implementation of it as well or something else

Yeah, that must be the reason. Hm, not sure, maybe it's possible to check in BXT when registering cvar if it's already registered, and in that case bind to the existing cvar value? Idk if that's easy to do with the current BXT code but that would be a better solution I think

PR #415, opened 2023-02-06
More bxt_ch_ additions

SmileyAG:
# Improved

- bxt_print_entities

# Added
- bxt_ch_teleport_to_entity
- bxt_ch_get_other_player_info
- bxt_ch_get_velocity
- bxt_get_servertime
- bxt_print_entities_by_index
- bxt_ch_entity_set_health
- bxt_ch_monster_set_origin

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -2871,102 +2871,38 @@ struct HwDLL::Cmd_BXT_CH_Get_Velocity
 	{
 		auto &hw = HwDLL::GetInstance();
 		const auto& vel = (*hw.sv_player)->v.velocity;
-		hw.ORIG_Con_Printf("Velocity: X = %f; Y = %f; Z = %f\n", vel.x, vel.y, vel.z);
+		hw.ORIG_Con_Printf("bxt_ch_set_vel %f %f %f\n", vel.x, vel.y, vel.z);
 		hw.ORIG_Con_Printf("Velocity (XY): %f\n", vel.Length2D());
 		hw.ORIG_Con_Printf("Velocity (XYZ): %f\n", vel.Length());
 	}
 };
 
-struct HwDLL::Cmd_BXT_CH_Get_BaseVelocity
+struct HwDLL::Cmd_BXT_CH_Get_Other_Player_Info
 {
 	NO_USAGE();
 
 	static void handler()
 	{
 		auto &hw = HwDLL::GetInstance();
+		auto &cl = ClientDLL::GetInstance();
+
+		const auto& mvtype = (*hw.sv_player)->v.movetype;
 		const auto& basevel = (*hw.sv_player)->v.basevelocity;
+		const auto& punch = (*hw.sv_player)->v.punchangle;
+
+		if (cl.pEngfuncs)
+			hw.ORIG_Con_Printf("Client maxspeed: %f\n", cl.pEngfuncs->GetClientMaxspeed());
+		hw.ORIG_Con_Printf("Movetype: %d (%s)\n", mvtype, hw.GetMovetypeName(mvtype));
+		hw.ORIG_Con_Printf("Health: %f\n", (*hw.sv_player)->v.health);
+		hw.ORIG_Con_Printf("Armor: %f\n", (*hw.sv_player)->v.armorvalue);
+		hw.ORIG_Con_Printf("Waterlevel: %d\n", (*hw.sv_player)->v.waterlevel);
+		hw.ORIG_Con_Printf("Max health: %f\n", (*hw.sv_player)->v.max_health);
+		hw.ORIG_Con_Printf("Gravity: %f\n", (*hw.sv_player)->v.gravity);
+		hw.ORIG_Con_Printf("Friction: %f\n", (*hw.sv_player)->v.friction);
 		hw.ORIG_Con_Printf("Basevelocity: X = %f; Y = %f; Z = %f\n", basevel.x, basevel.y, basevel.z);
 		hw.ORIG_Con_Printf("Basevelocity (XY): %f\n", basevel.Length2D());
 		hw.ORIG_Con_Printf("Basevelocity (XYZ): %f\n", basevel.Length());
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_Movetype
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
-		const auto& mvtype = (*hw.sv_player)->v.movetype;
-		hw.ORIG_Con_Printf("Movetype: %d (%s)\n", mvtype, hw.GetMovetypeName(mvtype));
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_MaxHealth
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
-		hw.ORIG_Con_Printf("Max health: %f\n", (*hw.sv_player)->v.max_health);
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_Health_And_Armor
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
-		hw.ORIG_Con_Printf("Health: %f; Armor: %f\n", (*hw.sv_player)->v.health, (*hw.sv_player)->v.armorvalue);
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_Waterlevel
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
-		hw.ORIG_Con_Printf("Waterlevel: %d\n", (*hw.sv_player)->v.waterlevel);
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_PunchAngle
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
-		const auto& punch = (*hw.sv_player)->v.punchangle;
-		hw.ORIG_Con_Printf("Punchangle: X = %f; Y = %f; Z = %f\n", punch.x, punch.y, punch.z);
-	}
-};
-
-struct HwDLL::Cmd_BXT_CH_Get_Gravity
-{
-	NO_USAGE();
-
-	static void handler()
-	{
-		auto &hw = HwDLL::GetInstance();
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -2837,6 +2837,82 @@ struct HwDLL::Cmd_BXT_Set_Angles
...
+		if ((*hw.sv_player)->v.flags & FL_ONGROUND)
+			out << "FL_ONGROUND; ";
+		if ((*hw.sv_player)->v.flags & FL_WATERJUMP)
+			out << "FL_WATERJUMP; ";
+		if ((*hw.sv_player)->v.flags & FL_FROZEN)
+			out << "FL_FROZEN; ";
+		if ((*hw.sv_player)->v.flags & FL_DUCKING)
+			out << "FL_DUCKING; ";
+		if ((*hw.sv_player)->v.flags & FL_ONTRAIN)
+			out << "FL_ONTRAIN";
```
YaLTeR, line 2896:
```suggestion
			out << "FL_ONTRAIN; ";
```
For consistency.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3765,6 +3933,48 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+			if (ent->v.solid == SOLID_BSP || ent->v.movetype == MOVETYPE_PUSHSTEP || is_trigger || is_ladder || is_friction || is_water)
+				origin = ent->v.origin + ((ent->v.mins + ent->v.maxs) / 2.f);
+			else
+				origin = ent->v.origin;
+
+			out << "bxt_ch_set_pos " << origin.x << " " << origin.y << " " << origin.z;
+
+			out << '\n';
+		}
+
```
YaLTeR, line 3972: If the entity was not found it should print an error.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3765,6 +3933,48 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+		std::ostringstream out;
+
+		edict_t *edicts;
+		const int numEdicts = hw.GetEdicts(&edicts);
+		for (int e = 0; e < numEdicts; ++e) {
+			const edict_t *ent = edicts + e;
+			if (!hw.IsValidEdict(ent))
+				continue;
+
+			if (e != num)
```
YaLTeR, line 3953: Why does this need to be a loop? Just check if `num < numEdicts` and that it's a valid ent?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3750,13 +3825,106 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+			bool is_friction = std::strncmp(classname, "func_friction", 13) == 0;
+			bool is_water = std::strncmp(classname, "func_water", 10) == 0;
+
+			Vector origin;
+			if (ent->v.solid == SOLID_BSP || ent->v.movetype == MOVETYPE_PUSHSTEP || is_trigger || is_ladder || is_friction || is_water)
+				origin = ent->v.origin + ((ent->v.mins + ent->v.maxs) / 2.f);
+			else
+				origin = ent->v.origin;
+
+			out << "; xyz: " << origin.x << " " << origin.y << " " << origin.z;
```
YaLTeR, line 3926:
```suggestion
			out << "; position: " << origin.x << " " << origin.y << " " << origin.z;
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3750,13 +3825,106 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+
```
YaLTeR, line 3927: Please extract this stuff into a separate helper function and call it from both handlers.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3736,7 +3811,7 @@ struct HwDLL::Cmd_BXT_Print_Entities
 {
 	NO_USAGE();
```
YaLTeR, line 3812: With two commands with arguments this now needs a usage help string.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3750,13 +3825,106 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+			else if (std::strcmp(name2, "strcmp") == 0)
+			{
+				if ((std::strcmp(classname, name1) != 0))
+					continue;
+			}
+			else
+			{
+				if ((strstr(classname, name1) == 0) && (strstr(classname, name2) == 0))
+					continue;
+			}
```
YaLTeR, line 3844: This is quite cursed. Honestly I'm having second thoughts on whether this two-argument variant is even needed in the first place. I'd rather just remove it

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3750,13 +3825,106 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+
+		edict_t *edicts;
+		const int numEdicts = hw.GetEdicts(&edicts);
+		for (int e = 0; e < numEdicts; ++e) {
+			const edict_t *ent = edicts + e;
+			if (!hw.IsValidEdict(ent))
+				continue;
+			
+			const char *classname = hw.GetString(ent->v.classname);
+			if (strstr(classname, name) == 0)
```
YaLTeR, line 3895: Could you add a second check here for targetname, if it exists? So both classname and targetname strstr will always result in printing, I think that would make sense

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4632,6 +4845,7 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 		Cmd_BXT_Set_Angles,
 		Handler<float, float>,
 		Handler<float, float, float>>("bxt_set_angles");
+	wrapper::Add<Cmd_BXT_Get_ServerTime, Handler<>>("bxt_get_servertime");
```
YaLTeR, line 4848:
```suggestion
	wrapper::Add<Cmd_BXT_Get_Server_Time, Handler<>>("bxt_get_server_time");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5648,6 +5861,26 @@ void HwDLL::KeyUp(Key& key)
...
+	switch (moveType)
+	{
+		case MOVETYPE_NONE:             return "None";
+		case MOVETYPE_WALK:             return "Walk";
+		case MOVETYPE_STEP:             return "Step";
+		case MOVETYPE_FLY:              return "Fly";
+		case MOVETYPE_TOSS:             return "Toss";
+		case MOVETYPE_PUSH:             return "Push";
+		case MOVETYPE_NOCLIP:           return "Noclip";
+		case MOVETYPE_FLYMISSILE:       return "Fly-missle";
```
YaLTeR, line 5875:
```suggestion
		case MOVETYPE_FLYMISSILE:       return "Fly-missile";
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5648,6 +5861,26 @@ void HwDLL::KeyUp(Key& key)
...
+		case MOVETYPE_BOUNCE:           return "Bounce";
+		case MOVETYPE_BOUNCEMISSILE:    return "Bounce-missle";
```
YaLTeR, line 5877:
```suggestion
		case MOVETYPE_BOUNCEMISSILE:    return "Bounce-missile";
```

YaLTeR (changes requested):

BunnymodXT/modules/HwDLL.cpp
SmileyAG: Don't want to make output longer to not cause issues with printing more symbols at once, let's keep it as `xyz:` instead

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3758,16 +3833,67 @@ struct HwDLL::Cmd_BXT_FreeCam
...
-	NO_USAGE();
+	USAGE("Usage:\n"
+	"bxt_print_entities <classname>\n"
+	"bxt_print_entities <targetname>\n"
+	"bxt_print_entities <target>\n"
+	"bxt_print_entities <classname> <classname>\n"
+	"bxt_print_entities <classname> strcmp\n"
+	"bxt_print_entities <classname> strstr\n"
+	"bxt_print_entities <targetname> targetname\n"
+	"bxt_print_entities <target> targetname\n"
```
YaLTeR, line 3884: This is really not a good UI. I think we should just leave a no-argument (prints all entities) and the single-argument (prints entities matching classname or targetname) cases.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3758,16 +3833,48 @@ struct HwDLL::Cmd_BXT_FreeCam
 	}
 };
 
+void HwDLL::PrintEntities(std::ostringstream &out, int e, const edict_t* ent)
```
YaLTeR, line 3836:
```suggestion
void HwDLL::PrintEntity(std::ostringstream &out, int index)
```
It can get edict itself, especially since it already gets classname and whatnot.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3776,21 +3883,173 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+
+			HwDLL::GetInstance().PrintEntities(out, e, ent);
+		}
+
+		auto str = out.str();
+		hw.ORIG_Con_Printf("%s", str.c_str());
+	}
+};
+
+struct HwDLL::Cmd_BXT_Print_Entities_Help
```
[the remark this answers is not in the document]
SmileyAG: Sure.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3776,21 +3883,173 @@ struct HwDLL::Cmd_BXT_Print_Entities
...
+		out << "bxt_print_entities <target>\n";
+		out << "bxt_print_entities <classname> <classname>\n";
+		out << "bxt_print_entities <classname> *\n";
+
+		auto str = out.str();
+		hw.ORIG_Con_Printf("%s", str.c_str());
+	}
+};
+
+struct HwDLL::Cmd_BXT_Print_Entities_By_Index
```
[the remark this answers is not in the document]
SmileyAG: - If printed message is too long (e.g.: output from `bxt_print_entities`) then it will not be able to be fully receive and as a result, half of the text will not be printed, so that we can split it with `bxt_print_entities_by_index`

- Example I can get index of needed entity under crosshair from `bxt_hud_entity_info`, and then I use it for `bxt_print_entities_by_index` to print needed data about entity and use for further research

YaLTeR:
Thanks!

PR #421, opened 2023-02-13
rpc: removed extra definition for chapter of the same map (ba_security1)

SmileyAG:
It didn't affect anything even without of that change, when I made those lists I did a lot of CTRL+C and CTRL+V and I didn't noticed that I wrote the same map twice, I only noticed this yesterday when used map lists for something in `discord_integration.hpp` lmao


YaLTeR:
Thanks

PR #422, opened 2023-02-16
Add some comments in engine headers/BXT files and hooked 'engineapi' for oldest WON builds

SmileyAG:
- Added comments in some of engine headers to show in which builds the struct ends on a specific function
- Hooked `engine_api_t` interface that removed starting from 3248 build or earlier, but at least exist in 1712 build (could be used  later for replacing pattern hooks of some functions with the calling functions from the structs itself)


HLSDK/engine/eiface.h
```diff
@@ -252,7 +252,7 @@ typedef struct enginefuncs_s
 
 	void		(*pfnGetPlayerStats)		( const edict_t *pClient, int *ping, int *packet_loss );
 
-	void		(*pfnAddServerCommand)		( const char *cmd_name, void (*function) (void) );
+	void		(*pfnAddServerCommand)		( const char *cmd_name, void (*function) (void) ); // Smiley: Last function in 1712 build
```
YaLTeR, line 255: Could you change these comments to be below (so like in-between)? Something like:

```
void (*old_function)();

// Functions below are present only in builds > 1712

void (*new_function)();
```

HLSDK/common/engine_launcher_api.h
```diff
@@ -0,0 +1,101 @@
+// engine/launcher interface
```
YaLTeR, line 1: Where does this file come from?

SmileyAG: That header is exist in HLSDK 2.0-HLSDK 2.3, but got removed with release SDK to GitHub

YaLTeR:
Thanks

PR #424, opened 2023-02-17
Add 'BXT_DISABLE_VSYNC' environment variable

SmileyAG:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4613,6 +4629,7 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 		CmdFuncs::AddCommand("notarget", ORIG_Host_Notarget_f);
 	}
 
+	wrapper::Add<Cmd_BXT_Disable_V_Sync, Handler<>>("bxt_disable_v_sync");
```
YaLTeR, line 4632:
```suggestion
	wrapper::Add<Cmd_BXT_Disable_VSync, Handler<>>("bxt_disable_vsync");
```

YaLTeR (changes requested):
Does this also work on Steam builds?

SmileyAG:
> Does this also work on Steam builds?

It does, but it would overriden after changing value in `gl_vsync` cvar anyway, so I don't think it there any sense to use it in Steam.

README.md
```diff
@@ -35,6 +35,7 @@ Use the [Bunnymod XT Launcher](https://github.com/YaLTeR/bxt-launcher).
 - **SPTLIB_DEBUG** - if set to 1, logs all dlopen, dlclose and dlsym calls.
 - **BXT_DISABLE_DEBUG_CONSOLE** - if set, disables the Bunnymod XT debug console.
 - **BXT_DISABLE_DISCORD_RPC** - if set, disables the Discord RPC module.
+- **BXT_DISABLE_VSYNC** - if set, disables the V-Sync on initialization.
```
YaLTeR, line 38:
```suggestion
- **BXT_DISABLE_VSYNC** - if set, disables the V-Sync on initialization (Windows-only).
```

YaLTeR:
Thanks

PR #427, opened 2023-03-01
Minor tweaks

SmileyAG:


BunnymodXT/patterns.hpp
```diff
@@ -118,7 +118,7 @@ namespace patterns
 			"55 8B EC 51 8B 45 ?? 50 E8 ?? ?? ?? ?? 83 C4 04 0F BE 08 85 C9 74 16 8B 55 08 52 68 ?? ?? ?? ?? E8 ?? ?? ?? ?? 83 C4 08"
 		);
 
-		PATTERNS(Host_AutoSave_f,
+		PATTERNS(Host_ValidSave,
```
YaLTeR, line 121: I thought it was called AutoSave? Or is it an inlined function?

SmileyAG: `Host_ValidSave` calling in **Huto_AutoSave_f** and **Host_Savegame_f** functions on Windows

In `hw.so` (Linux engine) it becomes as inline so that why you don't see calls to `Host_ValidSave` in Ghidra from those functions

YaLTeR:
Thanks

PR #428, opened 2023-03-01
com_model: fixed structs, now it same as in the GoldSrc engine (hw)

SmileyAG:
Reversed it from `hw.so` of 6153 build and by looking to `gl_model.h` from Quake 1

### DIFF
--- a/HLSDK/common/com_model.h
+++ b/HLSDK/common/com_model.h
@@ -94,12 +94,14 @@ typedef struct texture_s
 {
 	char		name[16];
 	unsigned	width, height;
+	int			gl_texturenum;
+	struct msurface_s	*texturechain;	// for gl_texsort drawing
 	int			anim_total;				// total tenths in sequence ( 0 = no)
 	int			anim_min, anim_max;		// time for this frame min <=time< max
 	struct texture_s *anim_next;		// in the animation sequence
 	struct texture_s *alternate_anims;	// bmodels in frame 1 use these
 	unsigned	offsets[MIPLEVELS];		// four mip maps stored
-	unsigned	paloffset;
+	byte	*pPal;
 } texture_t;
 
 typedef struct
@@ -118,7 +120,7 @@ typedef struct mnode_s
 	int			contents;		// 0, to differentiate from leafs
 	int			visframe;		// node needs to be traversed if current
 	
-	short		minmaxs[6];		// for bounding box culling
+	float		minmaxs[6];		// for bounding box culling
 
 	struct mnode_s	*parent;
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
How's it possible that the file from HLSDK is incorrect? Wouldn't that break any engine function that deals with the type?

SmileyAG:
> How's it possible that the file from HLSDK is incorrect? Wouldn't that break any engine function that deals with the type?

# Start

- Quake and Half-Life having two types of engine: **Hardware** and **Software** (hw.dll and sw.dll)
- Structures between these types of engine is differ
- Valve included in HLSDK structs to `com_model.h` for **Software** engine
- BunnymodXT is running on **Hardware** engine, so that's the reason why it needs to be fixed


For confirmation could be opened `hw.so` in Ghidra, type name of struct that differ on other engine (e.g. `texture_`) in **Data Type Manager** and look to inside of them:



Structs from `GL_MODEL.h` - **Hardware** engine
Structs from `model.h` - **Software** engine (structs from that header is nowhere used for hw.so/hw.dll, it just exists in it for unknown reason)

# Examples

## GL_MODEL.h (`texture_t`)



## model.h (`texture_t`)



# End

If this is not enough, you can compare these two headers from Quake code: 

https://github.com/id-Software/Quake/blob/master/WinQuake/gl_model.h
https://github.com/id-Software/Quake/blob/master/WinQuake/model.h

YaLTeR:
Thanks, interesting. Weird though, are there no API calls which expose any of the conflicting structures? Because then wouldn't all mods be broken? Since everyone runs in hardware engine.

SmileyAG:
> Thanks, interesting. Weird though, are there no API calls which expose any of the conflicting structures?

No idea, honestly.

> Because then wouldn't all mods be broken? Since everyone runs in hardware engine.

That would break only in the case if:

- Modder used structs for Software engine in HLSDK, while we would running game in Hardware
- Vice-versa (modification is uses structs for Hardware engine, while we would running game in Software)

In the HLSDK from what I looked at, only `model_t`/`model_s` is used in functions, but I will answer you the reason why everything continues to work on both engines: that struct is having the same size (0x188) and same variables between the engines

Of course that it you would call variables from other struct variables of `model_s` then it would break one of the engine (e.g.: `pModel->surfaces->dlightframe`)

But at least from what I looked up in HLSDK, I didn't found cases of that.

So if you want to support BXT for the Software engine, then we will have to compile a separate versions (due to the use of `#ifdef` in the headers), but I honestly don’t see much point in this. Majority of peoples playing in Hardware engine, so if modders want to utilize them in code they should definitely to fix structs at first.

YaLTeR:
I see, thanks again!

PR #429, opened 2023-03-05
CMake: added 'SOFT' flag to compile with compatible structures for software engine

SmileyAG:


YaLTeR:
Thanks. Guess it needs to be added to the CI too (but probably not to releases)

PR #432, opened 2023-03-06
Automate the process of finding engine offsets

SmileyAG:
This is useful because the offsets for engine builds can be different and hardcoding them is a very questionable solution.


YaLTeR:
Thanks

PR #435, opened 2023-03-08
Permanently reverted 'bxt_hud_game_alpha' back and add 'bxt_hud_game_alpha_damage'

SmileyAG:
Talked with friend about rendering runs, and we got into conclusion that `bxt_hud_game_alpha` should be returned in replace of `bxt_hud_game_alpha_max_clientside`

If you have a question what is that specific code in `bxt_hud_game_alpha_damage`: https://github.com/ValveSoftware/halflife/blob/master/cl_dll/health.cpp#L377
That makes smoothly fading HUD damage icons from alpha that specified in **bxt_hud_game_alpha** to `0` (cuz by default it is hardcoded to max value aka `256.0`)

If you do not understand what are these damage icons that are talking about:

изображение

### DIFF
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -114,7 +114,7 @@
 	X(bxt_unlock_camera_during_pause, "0") \
 	X(bxt_hud, "1") \
 	X(bxt_hud_color, "") \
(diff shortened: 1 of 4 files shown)


YaLTeR:
Thanks!

PR #436, opened 2023-03-10
Fixed data types from 'bool' to 'qboolean' (int) for some variables

SmileyAG:
I must to say that it isn't caused any issues in-game for all this time, it just my curiosity which decided to double-check the BXT code for data types of variables, and then I remembered about `bool` and `qboolean` and their differences:

Definition of `qboolean` in C++ code is an **int** - which weights is `4` bytes, while `bool` from C++ equals to `1` byte

Proof that those variables is `qboolean` (int):

`noclip_anglehack` = https://github.com/id-Software/Quake/blob/bf4ac424ce754894ac8f1dae6a3981954bc9852d/WinQuake/host_cmd.c#L149 (that could be also confirmed from debug information of `hw.so` aka GoldSrc engine)
`bInDuck` = https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/pm_shared/pm_defs.h#L110

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -1618,7 +1618,7 @@ void HwDLL::FindStuff()
 					svs = reinterpret_cast<svs_t*>(*reinterpret_cast<uintptr_t*>(f + 79) - 8);
 					offEdict = *reinterpret_cast<ptrdiff_t*>(f + 182);
 					offActiveAddr = *reinterpret_cast<uintptr_t*>(f + 0x32);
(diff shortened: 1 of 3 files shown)


YaLTeR:
Ugh, didn't realize it would create a merge commit... I guess we'll have to live with it now, unless you'd be willing to rebase to get rid of it

SmileyAG:
> Ugh, didn't realize it would create a merge commit... I guess we'll have to live with it now, unless you'd be willing to rebase to get rid of it

It is rebased now.

YaLTeR:
Thanks!

PR #442, opened 2023-04-30
Add bxt_tas_editor_show_from_last_frames

khanghugo:
Camera lines still show up normally.

### DIFF
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -38,6 +38,7 @@
 	X(bxt_tas_editor_apply_smoothing_over_s, "0.15") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_duration, "0.03") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_multiplier, "3") \
+	X(bxt_tas_editor_show_from_last_frames, "0") \
 	X(bxt_tas_norefresh_until_last_frames, "0") \
 	X(bxt_tas_write_log, "0") \
 	X(bxt_tas_playback_speed, "1") \
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -38,7 +38,7 @@
 	X(bxt_tas_editor_apply_smoothing_over_s, "0.15") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_duration, "0.03") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_multiplier, "3") \
-	X(bxt_tas_editor_show_from_last_frames, "0") \
+	X(bxt_tas_editor_show_only_last_frames, "0") \
 	X(bxt_tas_norefresh_until_last_frames, "0") \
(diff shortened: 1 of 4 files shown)


khanghugo:
I think camera should be cut off as well. Let me make that change.

khanghugo:
This should be good. One weird thing with camera is that stationary green line will show up as blue if segment is too short, but it is green underneath and can be smoothed just fine. That's more of the drawing and it is ignored.

BunnymodXT/triangle_drawing.cpp
```diff
@@ -1610,6 +1623,13 @@ namespace TriangleDrawing
 			Vector last_shown_view_angle_origin;
 
 			for (size_t frame = 1; frame < player_datas.size(); ++frame) {
+				if (frame < start_frame) {
+					// Match inner advancing when skip showing some frame bulks
```
YaLTeR, line 1627: I don't quite understand this comment, could you rephrase it?

BunnymodXT/cvars.hpp
```diff
@@ -38,6 +38,7 @@
 	X(bxt_tas_editor_apply_smoothing_over_s, "0.15") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_duration, "0.03") \
 	X(_bxt_tas_editor_apply_smoothing_high_weight_multiplier, "3") \
+	X(bxt_tas_editor_show_from_last_frames, "0") \
```
YaLTeR, line 41: show_only_last_frames perhaps

BunnymodXT/triangle_drawing.cpp
khanghugo: https://github.com/YaLTeR/BunnymodXT/blob/6efb3f65bcbf6a97d06c2550182d84f1ed085b58/BunnymodXT/triangle_drawing.cpp#L1750

From here, it draws perpendicular lines and some basic interactions. Due to how everything is structured, if I don't change `next_frame_bulk_start_index` when skipping showing the current frame (meaning not going over that line), there will be no perpendicular lines and basic interactions.

khanghugo: "Incrementing `next_frame_bulk_start_index` when frames are skipped to correctly render perpendicular line."

BunnymodXT/triangle_drawing.cpp
```diff
@@ -592,6 +592,11 @@ namespace TriangleDrawing
 		const auto& normalzs = input.normalzs;
 		const auto& frame_bulk_starts = input.frame_bulk_starts;
 
+		const auto show_only_last_frames = CVars::bxt_tas_editor_show_only_last_frames.GetInt();
+		const auto start_frame = show_only_last_frames <= 0 || 
+			(unsigned) show_only_last_frames >= input.player_datas.size() ? 
+				1 : input.player_datas.size() - show_only_last_frames;
```
YaLTeR, line 598: Could you rewrite this as a normal if statement? Seems too complicated for a ternary.

YaLTeR:
Thanks

PR #443, opened 2023-05-13
The beginning of the end?

SmileyAG:
SPOILER ALERT: that request might contains dubious code, unfinished implementations, incomplete description of changes here, potential crashes and some extra hooks of functions. That's still work-in-progress, don't take it too serious until that pull request will goes out of draft stage.

Erm, where should I start? Well, most of the changes I did about a month ago, but I didn't requested it to here even as a draft, because, yeah.

After that request would be done out of draft stage, it would also need to add new builds into the CI file, so we will get the next:

```
Linux (Debug)
Linux (Release)
Windows (Debug)
Windows (Release)
Windows, CoF (Debug)
Windows, CoF (Release)
Windows, HLSDK 1.0 (Debug)
Windows, HLSDK 1.0 (Release)
```

Let me list the changes for you so that you have an overall picture for this:

# bxt_autojump_fix

- Modifies **m_afButtonPressed** at the beginning of `CBasePlayer::Jump` to pass the function if autojump enabled.

# bxt_disable_cheats_check_in_demo

- Modifies **cls.state** in `Cvar_Command` / `Cvar_CommandWithPrivilegeCheck` functions and not returning the `R_ForceCvars` to allow enabling the cheat commands for the demos that recorded on multiplayer servers.

# bxt_ch_get_steamid_in_demo

- Allows to get the STEAMID of player from the demo that recorded on **4554** engine build or newer.

# Lot of new and old enhanced commands that allowing to control over the player variables, teleporting to them or get full information

```
bxt_ch_client_set_armor
bxt_ch_client_set_velocity
bxt_ch_set_ammo_clip
bxt_ch_set_ammo_secondary
bxt_ch_set_ammo_primary
bxt_ch_get_other_player_info
bxt_ch_monster_set_origin
bxt_ch_entity_set_health
bxt_ch_teleport_to_entity
bxt_ch_client_toggle_noclip
```

# Fixed crash when the non-host players with the BXT injected on the server starts to demo record

This was due to #293 request, in which I'm getting the player health value ​​from `sv_player` variable, but the client cannot access it and due to this the crash is occurs.

# Experimental support for Sven Co-op 5.25 (latest version)

TODO

# Experimental support for HLSDK 1.0 builds (pre-1.1.0.0 WON versions)

I want to highlight the fact that example the functions from `pm_shared.c` are implemented in the engine on pre-1.1.0.0 WON versions or there is also `SV_AddToFullPack` function instead of `AddToFullPack` from the server-side, I got found it by reversing **HLDS 3.0.1.6 Linux** version (file: `hlds_l`)

Downloaded from here: https://files.ndlp.info/cstrike/linux/hlds/updates/hlds_lupgrd3016.tar.gz


khanghugo:
Nice work. Since you have some interests in cheat commands allowing player to traverse easier, I think something toggles like noclip toggle or "hook" toggle built in would be very nice. If it doesn't fit the scope of this PR, I guess I can find sometimes and add it seperately.

SmileyAG:
> Nice work. Since you have some interests in cheat commands allowing player to traverse easier, I think something toggles like noclip toggle or "hook" toggle built in would be very nice. If it doesn't fit the scope of this PR, I guess I can find sometimes and add it seperately.

Hey! Well, yeah, my initial goal for creating a new commands that allow you to change the player variables is for the order to practice speedrun routes with friends in co-op or rather just to have fun.

As for the cmd code to toggling the noclip of players, the implementation should be very simple:

```
if (ent->v.movetype == MOVETYPE_NOCLIP)
	ent->v.movetype = MOVETYPE_WALK;
else
	ent->v.movetype = MOVETYPE_NOCLIP;
```

khanghugo:
For the hook toggle, it can be something like this: https://youtu.be/m3pSAElq_lc

These two commands are used a lot in KZ servers and they also have another extra toggle while pressed to go faster.

SmileyAG:
Well, I can say that I will probably make one more commit to this branch and then I'll close that single big request and instead do a several ones with more competent sorting of commits

Want to say that this request is more like a small warehouse of my good and bad decisions, most of things would be implemented later, but some of them might not added at all (since they are likely requires a deep reworking, so I might not finish it in the case if I abandon all related to HL or something terrible will happen to me), I'll just leave that stuff from which you can find out something for future, e.g. support of some features for old versions of the GoldSrc engine

In addition, for that request I found offsets to variables even for those versions of the engine in which many things can already be accessed through structures, but I am sure that with the next pull requests I will try to reduce the number of extra hooks and find variables only for those versions where it can used here and now (this also leads to the idea that I will probably remove some of the patterns in the client / server sides for those functions where it does not carry any support for the functionality, because from due to the large number of patterns and their length, there are cases when in some mods they do a false accusations, but that removing would be probably a slightly, so it will not affect the most of stuff)

I also want to take the time to double-check once and for all the offsets / patterns to be ensure that there is no wrong finds

YaLTeR:
Alright, sounds great.

khanghugo:
Are you alright? That first part sounds a bit grim.

SmileyAG:
Decided to finally closing that in considering to split it for multiple requests

Months ago thought that I could handle of managing it, oh boy, that was a wrong decision, since it slowly started to be at huge scale than it was looked at initial

I'm got for now a free time and some spirit to deal with it again, so the first chunks would be requested today and continue upon next few weeks

From now features would be a have a definitely contain a some comments with explanations, since I can see that someone might to be confused about decisions that done
Also, starting for a while I'm setupped machine with Ubuntu on it, so I would definitely re-look later to cases if some features is supported only on Windows builds atm

I'm even considering it might be worth to write and request a some small documentation in favor of new and might be even for old contributors at some point?
Because I can see that most of peoples (including me old one) don't care much about compatibility across most of versions, so one example of that is the current code relying too much on interface functions (e.g.: pEngfuncs->pTriAPI), while I see that it definitely requires to make for them a wrapper functions, so that would possible to make support it across of all versions, due of having a different code per version inside of one function

YaLTeR:
Glad to see you feeling better :)

> Months ago thought that I could handle of managing it, oh boy, that was a wrong decision, since it slowly started to be at huge scale than it was looked at initial

Heh, yep. Then imagine reviewing it.

> From now features would be a have a definitely contain a some comments with explanations, since I can see that someone might to be confused about decisions that done

Sounds great.

> I'm even considering it might be worth to write and request a some small documentation in favor of new and might be even for old contributors at some point?

Yeah, that's a very good idea.

> Because I can see that most of peoples (including me old one) don't care much about compatibility across most of versions, so one example of that is the current code relying too much on interface functions (e.g.: pEngfuncs->pTriAPI), while I see that it definitely requires to make for them a wrapper functions, so that would possible to make support it across of all versions, due of having a different code per version inside of one function

I didn't know better when I started BXT, and then it spiraled out of control, so yeah. In bxt-rs I strive to do it right.

PR #446, opened 2023-06-23
Add runtime data parsing and demo accurate timer

### LINKED ISSUE #294: Parse health from BXT runtime of current demo to use for "bxt_hud_health" command
I already logged player health to demos in #293 request, but I'm kinda 2Head in parsing from it, so, implement it please if u get free time, @YaLTeR

eim64:
Adds 2 hw.dll hooks:
- ValidStuffText 
- CL_ReadDemoMessage_OLD

That are used to read the //BXTD0 commands and corresponding runtime data each frame. These have only been tested on latest steam version of Half-Life.

Conversion back to a vector<Data> was implemented by adding a decrypt filter and load template.
demo_data_visitor is then applied to the recovered data and should contain logic such as updating timer/health. This could arguably be part of the load template, but imo it's better for parse logic to be separate. 

I currently set HwDll::executing to true whenever a demo frame is read, to make the timer resume after save/loads. This is scuffed, and the Cbuf_Execute hook should be fixed to resume in demos instead, but I dont know how to.

Can be used to resolve #294

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -317,6 +317,16 @@ extern "C" qboolean __cdecl CL_CheckGameDirectory(char *gamedir)
 {
 	return HwDLL::HOOKED_CL_CheckGameDirectory(gamedir);
 }
+
+extern "C" int __cdecl ValidStuffText(char *buf)
+{
+	return HwDLL::HOOKED_ValidStuffText(buf);
+}
+
+extern "C" qboolean __cdecl CL_ReadDemoMessage_OLD()
+{
+	return HwDLL::HOOKED_CL_ReadDemoMessage_OLD();
+}
 #endif
 
 void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* moduleBase, size_t moduleLength, bool needToIntercept)
@@ -439,6 +449,8 @@ void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* modul
 			MemUtils::MarkAsExecutable(ORIG_SCR_NetGraph);
 			MemUtils::MarkAsExecutable(ORIG_Host_Shutdown);
 			MemUtils::MarkAsExecutable(ORIG_ReleaseEntityDlls);
+			MemUtils::MarkAsExecutable(ORIG_ValidStuffText);
+			MemUtils::MarkAsExecutable(ORIG_CL_ReadDemoMessage_OLD);
 		}
 
 		MemUtils::Intercept(moduleName,
@@ -499,6 +511,8 @@ void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* modul
 			ORIG_CL_CheckGameDirectory, HOOKED_CL_CheckGameDirectory,
 			ORIG_SaveGameSlot, HOOKED_SaveGameSlot,
 			ORIG_ReleaseEntityDlls, HOOKED_ReleaseEntityDlls,
+			ORIG_ValidStuffText, HOOKED_ValidStuffText,
+			ORIG_CL_ReadDemoMessage_OLD, HOOKED_CL_ReadDemoMessage_OLD,
 			ORIG_Host_Shutdown, HOOKED_Host_Shutdown);
 	}
 
@@ -581,6 +595,8 @@ void HwDLL::Unhook()
 			ORIG_CL_CheckGameDirectory,
 			ORIG_SaveGameSlot,
 			ORIG_ReleaseEntityDlls,
+			ORIG_ValidStuffText,
+			ORIG_CL_ReadDemoMessage_OLD,
 			ORIG_Host_Shutdown);
 	}
 
@@ -679,6 +695,8 @@ void HwDLL::Clear()
 	ORIG_SCR_DrawPause = nullptr;
 	ORIG_Host_Shutdown = nullptr;
 	ORIG_ReleaseEntityDlls = nullptr;
+	ORIG_ValidStuffText = nullptr;
+	ORIG_CL_ReadDemoMessage_OLD = nullptr;
 
 	ClientDLL::GetInstance().pEngfuncs = nullptr;
 	ServerDLL::GetInstance().pEngfuncs = nullptr;
@@ -1013,6 +1031,18 @@ void HwDLL::FindStuff()
 		else
 			EngineDevWarning("[hw dll] Could not find CL_CheckGameDirectory.\n");
 
+		ORIG_ValidStuffText = reinterpret_cast<_ValidStuffText>(MemUtils::GetSymbolAddress(m_Handle, "ValidStuffText"));
+		if (ORIG_ValidStuffText)
+			EngineDevMsg("[hw dll] Found ValidStuffText at %p.\n", ORIG_ValidStuffText);
+		else
+			EngineDevWarning("[hw dll] Could not find ValidStuffText.\n");
+
+		ORIG_CL_ReadDemoMessage_OLD = reinterpret_cast<_CL_ReadDemoMessage_OLD>(MemUtils::GetSymbolAddress(m_Handle, "CL_ReadDemoMessage_OLD"));
+		if (ORIG_CL_ReadDemoMessage_OLD)
+			EngineDevMsg("[hw dll] Found CL_ReadDemoMessage_OLD at %p.\n", ORIG_CL_ReadDemoMessage_OLD);
+		else
+			EngineDevWarning("[hw dll] Could not find CL_ReadDemoMessage_OLD.\n");
+
 		if (!cls || !psv || !svs || !svmove || !ppmove || !host_client || !sv_player || !sv_areanodes || !cmd_text || !cmd_alias || !host_frametime || !cvar_vars || !movevars || !ORIG_SV_AddLinksToPM || !ORIG_SV_SetMoveVars)
 			ORIG_Cbuf_Execute = nullptr;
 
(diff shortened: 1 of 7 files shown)


BunnymodXT/runtime_data.cpp
```diff
@@ -352,4 +505,52 @@ namespace RuntimeData
...
 		Clear();
 	}
+
+	// Visitor applied to each Data object found in demo.
+	class demo_data_visitor : public boost::static_visitor<>
+    {
+    public:
+        void operator()(const RuntimeData::Time& t) const {
+            CustomHud::SetTime((int)t.hours, (int)t.minutes, (int)t.seconds, t.remainder);
+            CustomHud::SetCountingTime(true);
```
chinese-soup, line 515: This causes the timer to e.g. never stop in demos where bxt_timer_stop has been run manually etc. (so even adding SetCountingTime to `false` for the `GameEndMarker` visitor solve this).

I'd say since the time is being read from the runtime data inside the demos and the time is only saved every 1 second there, let's not lie to the end-user, and only update it with the real data from the demo. 

But I am no @YaLTeR, just my two cents.

chinese-soup (commented):
Good stuff. 

Please add Linux hooks as well:

```diff
diff --git a/BunnymodXT/modules/HwDLL.cpp b/BunnymodXT/modules/HwDLL.cpp
index 852c73f..f683a01 100644
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -317,6 +317,16 @@ extern "C" qboolean __cdecl CL_CheckGameDirectory(char *gamedir)
 {
 	return HwDLL::HOOKED_CL_CheckGameDirectory(gamedir);
 }
+
+extern "C" int __cdecl ValidStuffText(char *buf)
+{
+	return HwDLL::HOOKED_ValidStuffText(buf);
+}
+
+extern "C" qboolean __cdecl CL_ReadDemoMessage_OLD()
+{
+	return HwDLL::HOOKED_CL_ReadDemoMessage_OLD();
+}
 #endif
 
 void HwDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* moduleBase, size_t moduleLength, bool needToIntercept)
@@ -1021,6 +1031,18 @@ void HwDLL::FindStuff()
 		else
 			EngineDevWarning("[hw dll] Could not find CL_CheckGameDirectory.\n");
 
+		ORIG_ValidStuffText = reinterpret_cast<_ValidStuffText>(MemUtils::GetSymbolAddress(m_Handle, "ValidStuffText"));
+		if (ORIG_ValidStuffText)
+			EngineDevMsg("[hw dll] Found ValidStuffText at %p.\n", ORIG_ValidStuffText);
+		else
+			EngineDevWarning("[hw dll] Could not find ValidStuffText.\n");
+
+		ORIG_CL_ReadDemoMessage_OLD = reinterpret_cast<_CL_ReadDemoMessage_OLD>(MemUtils::GetSymbolAddress(m_Handle, "CL_ReadDemoMessage_OLD"));
+		if (ORIG_CL_ReadDemoMessage_OLD)
+			EngineDevMsg("[hw dll] Found CL_ReadDemoMessage_OLD at %p.\n", ORIG_CL_ReadDemoMessage_OLD);
+		else
+			EngineDevWarning("[hw dll] Could not find CL_ReadDemoMessage_OLD.\n");
+
 		if (!cls || !psv || !svs || !svmove || !ppmove || !host_client || !sv_player || !sv_areanodes || !cmd_text || !cmd_alias || !host_frametime || !cvar_vars || !movevars || !ORIG_SV_AddLinksToPM || !ORIG_SV_SetMoveVars)
 			ORIG_Cbuf_Execute = nullptr;

```

BunnymodXT/runtime_data.cpp
eim64: The `CustomHud::SetCountingTime(true);` line was a remnant from when I was trying to get the timer working after loads, and I simply forgot to remove it.

As for `bxt_timer_stop` and other bxt commands, they are not called when parsing the demo, but executing them based on runtime data seems to work.  I doubt it'll lead to unexpected behaviour, otherwise a more strict whitelist of commands are needed.

Pushed an update now.

chinese-soup: >was a remnant

Cool. :) Thanks!


>As for [...]

Yeah np, I really only meant specifically just removing the `SetCountingTime(true);` really, bxt_timer_stop was just an example.

BunnymodXT/runtime_data.cpp
```diff
@@ -506,4 +508,48 @@ namespace RuntimeData
 
 		Clear();
 	}
+
+	// Visitor applied to each Data object found in demo.
+	class demo_data_visitor : public boost::static_visitor<>
+    {
```
YaLTeR, line 514: The indentation is a bit all over the place here. Please fix it to be consistent with the rest of the file (so, tabs, and consistent indentation levels).

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -7482,3 +7516,36 @@ HOOK_DEF_0(HwDLL, void, __cdecl, ReleaseEntityDlls)
...
+HOOK_DEF_0(HwDLL, qboolean, __cdecl, CL_ReadDemoMessage_OLD)
+{
+	// CL_ReadDemoMessage_OLD calls ValidStuffText to check if the demo commands are valid
+	insideCL_ReadDemoMessage = true;
+	auto rv = ORIG_CL_ReadDemoMessage_OLD();
+	insideCL_ReadDemoMessage = false;
+
+	// Loads in demos do not set executing to true.
+	// This is a temporary and scuffed solution
+	executing = true;
```
YaLTeR, line 7545: Hmm, I suspect it's a bad idea to mess with `executing`, considering the TAS support relies heavily on it, and it's quite a spaghetti where I'm not sure myself what changing some variable like this can break. Semantically, `executing = true` is supposed to mean "this frame, player input will have an effect" (i.e. not be ignored during a load or something), which is not the case during demo playback.

But then again, if demo playback doesn't reset executing to false, and if this only affects playdemo which you are pretty much never doing when TAS playback is involved, maybe it's fine...

One question I have is, does this make it so bxt_timer_start will now work when playing back demos with no runtime data? If it didn't work before and will now work in this case, it might be undesired, because people might try to retime their demos that way thinking they have the accurate IGT when it isn't.

BunnymodXT/runtime_data.cpp
```diff
@@ -513,15 +513,28 @@ namespace RuntimeData
...
     {
     public:
-        void operator()(const RuntimeData::Time& t) const {}
+        void operator()(const RuntimeData::Time& t) const {
+            CustomHud::SetTime((int)t.hours, (int)t.minutes, (int)t.seconds, t.remainder);
+        }
+
+		void operator()(const RuntimeData::CommandExecution& e) const {
+			// This is to ensure commands like bxt_timer_stop are executed.
+			if(e.command.rfind("bxt_", 0) == 0) {
```
YaLTeR, line 522: Uhh, I don't think it's a good idea to run any bxt command from a demo. If you want bxt_timer_stop specifically (and any other specific commands), it's better to check for bxt_timer_stop exactly here.

BunnymodXT/runtime_data.cpp
```diff
@@ -513,15 +513,28 @@ namespace RuntimeData
...
+		}
+		
+        void operator()(const RuntimeData::GameEndMarker& m) const {
+			CustomHud::SetCountingTime(false);
+		}
+		
 		void operator()(const RuntimeData::PlayerHealth& p) const {}
         void operator()(const RuntimeData::VersionInfo& v) const {}
-        void operator()(const RuntimeData::CVarValues& v) const {}
+        void operator()(const RuntimeData::CVarValues& v) const { }
```
YaLTeR, line 534: Unnecessary space added.

BunnymodXT/runtime_data.cpp
YaLTeR: Same about changes to this file in the last commit.

YaLTeR (commented):
Hey! Sorry for taking so long to look at this, I'm just completely buried in finishing TAS editor 2.

This is indeed a valuable feature to have, and it mostly matches how I would implement it. I just have a few comments, and we need to decide what to do about `executing`.

BunnymodXT/modules/HwDLL.cpp
eim64: Previously, the timer during demo playback would start executing after map loads and then pause after save-loads. When I first wanted to record my runs I assumed the timer was accurate apart from these pauses. So it is a more valid concern, now that it runs smoothly. 

Setting executing to true after reading runtimedata would make it identical to its previous state for demos without runtime data. I could also set finishingLoad to true, if that makes more sense semantically, let me know if thats any better for tas interference.

BunnymodXT/runtime_data.cpp
eim64: Will add a whitelist of bxt commands instead 👍

BunnymodXT/modules/HwDLL.cpp
YaLTeR: I think/hope it should be fine as you made it, thanks

BunnymodXT/runtime_data.cpp
```diff
@@ -352,4 +508,69 @@ namespace RuntimeData
...
+	// Visitor applied to each Data object found in demo.
+	class demo_data_visitor : public boost::static_visitor<> {
+	public:
+		void operator()(const RuntimeData::Time& t) const {
+			CustomHud::SetTime((int)t.hours, (int)t.minutes, (int)t.seconds, t.remainder);
+		}
+
+		void operator()(const RuntimeData::CommandExecution& e) const {
+			for (const auto& cmd : cmd_whitelist) {
+				if (e.command.rfind(cmd, 0) == 0) {
```
YaLTeR, line 521: Why not if (e.command == cmd)?

eim64: I am checking for prefix in case there are spaces/newline at the end. I used a set at first but it didnt work, and assumed this to be the cause, but it could be wrong since i dont know how much text trimming the engine does.

YaLTeR: Could you check what the actual command is there? I suspect there might be a single \n or ; in most cases that you could trim manually.

YaLTeR: Hm, so as far as I can see, the commands are stored there as is, with no extra symbols.

YaLTeR: Well, I guess it's fine as is tbh let's leave it

PR #447, opened 2023-07-10
Add constant yawspeed

khanghugo:


khanghugo:
Close due to assigments being in hlstrafe instead.

Issue #341, opened 2022-10-05
WON viewroll (bxt_viewroll_angled?)


Droooga:

SmileyAG:
The only important thing I would like for that is to make it not override `viewangles` for the demo, otherwise not accepting that changes. (because that can cause a lot of issues, for example, moderators who verify the run, or if I want to render demo of someone who played with the viewroll but with a normal settings)

PR #452, opened 2023-07-30
Added CVars to restore WON view rolling effects

jcb936:
See issue in the original HL SDK repo: https://github.com/ValveSoftware/halflife/issues/1544

The viewroll effect was removed sometime when HL came to Steam. Since BunnymodXT can restore the weapon bob effect, I thought it would have been nice for it to also being able to restore the viewroll effect. The values are slightly different from the one in the issue above, and have been taken from the implementation used in halflife-updated (see: https://github.com/SamVanheer/halflife-updated/commit/dcd3e93a5abcf151c287c68d166953f8b400c7a5 ) which gives a nicer effect.
This PR adds cvars to restore the viewroll effect (defaults to false) and set the rollspeed and rollangle values.

### DIFF
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -210,6 +210,9 @@
 	X(bxt_viewmodel_ofs_right, "0") \
 	X(bxt_viewmodel_ofs_up, "0") \
 	X(bxt_viewmodel_bob_angled, "0") \
+	X(bxt_viewmodel_restore_viewroll, "0") \
+	X(bxt_viewmodel_viewrollangle, "2.0") \
+	X(bxt_viewmodel_viewrollspeed, "200") \
 	X(bxt_show_bullets, "0") \
 	X(bxt_show_bullets_enemy, "0") \
 	X(bxt_anglespeed_cap, "1") \
--- a/BunnymodXT/modules/ClientDLL.cpp
+++ b/BunnymodXT/modules/ClientDLL.cpp
@@ -973,6 +973,9 @@ void ClientDLL::RegisterCVarsAndCommands()
 		REG(bxt_viewmodel_ofs_up);
 		REG(bxt_viewmodel_bob_angled);
 		REG(bxt_remove_punchangles);
+		REG(bxt_viewmodel_restore_viewroll);
+		REG(bxt_viewmodel_viewrollangle);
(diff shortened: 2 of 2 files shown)


SmileyAG:
I suppose it doesn't override `viewangles` at the end and can be toggled to enable/disable in the demo? Otherwise, I would restrict it up to `sv_cheats 1` or banning for RTA runs (see https://github.com/YaLTeR/BunnymodXT/issues/341#issuecomment-1344644400)

Because, the known implementations of viewroll that I seen before was been overriding `viewangles`, that also saved up to the demos without possibilities for removing

SmileyAG:
> I suppose it doesn't override viewangles at the end and can be toggled to enable/disable in the demo? Otherwise, I would restrict it up to `sv_cheats 1` or banning for RTA runs (see [#341 (comment)](https://github.com/YaLTeR/BunnymodXT/issues/341#issuecomment-1344644400))
> 
> Because, the known implementations viewroll that I seen before was been overriding viewangles, that also saved up to the demos without possibilities for removing

Tested in-game and as I expected, that's still overrides view angles at the end, so it saves to the demo without a possibilities for toggling it
@YaLTeR Might be restrict it to `sv_cheats 1` then, what do you think?

YaLTeR:
Sure

jcb936:
Sadly it overrides the viewangles, yes! I can restrict it to sv_cheats 1 for now and in the meantime see if there's a more clever solution to not override the values for demo, would be nice to have.

khanghugo:
The mechanism to avoid writing overriden viewangles values and such into the demo migth be a bit complicated but shouldn't be impossible. I hypothesise you must have a static variable holding the current viewangles, then after something like `V_RenderView()` and between any calls with `cls.demorecording == true` condition, you can restore it. From a quick glance, I think it should be before `ClientDLL_UpdateClientData`.

jcb936:
Apologies was a bit busy, changed it to be available only with sv_cheats 1! Still looking into making it safe for demos and such.

khanghugo:
Is it necessary to add it? I mean it's nice if you figure that out because it involves demo and stuffs. Anyway, since it is only available under sv_cheats 1 now, maybe if you want you can go extra for that (which makes good foundation for other related functionalities to work with demo as well!).

YaLTeR:
Thanks!

PR #453, opened 2023-08-29
check to prevent ub

Eddio0141:


YaLTeR:
Thanks for finding and fixing this!

PR #458, opened 2023-10-27
Change some +bxt_ch_hook behaviors

### LINKED ISSUE #456: Correcting +bxt_ch_hook request
That a good addition, but I haven't looked to the request itself before, so let's see:

https://github.com/YaLTeR/BunnymodXT/blob/71f1c9a75d6e92cb1f5e48c8d380541451165ea8/BunnymodXT/modules/HwDLL.cpp#L3166

What is that random stuff does here? Please, don't add some wacky kind of that hardcoded stuff that ain't supposed to be in it, let player to set HP on their own with `bxt_ch_set_health` or using the `god` command

https://github.com/YaLTeR/BunnymodXT/blob/71f1c9a75d6e92cb1f5e48c8d380541451165ea8/BunnymodXT/modules/HwDLL.cpp#L3168

Don't make the speed is hardcoded, better to create the specified cvar for that initially, or otherwise soon or later you still would to hear from others for making that addition

Other than that, additionally, might to consider of adding draw of trace line with power of pTriAPI
Or rather re-using beam sprites from HL1 (e.g. from Gauss gun) and pEfxAPI/pEventAPI/pTriAPI/SPR_ functions to draw it

khanghugo:
Closes #456

No more HP thing, a cvar to control speed, and beam.

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -796,7 +796,6 @@ void HwDLL::Clear()
 	libTASExportFile.close();
 	ch_hook = false;
 	ch_hook_point = Vector();
-	ch_hook_hp_before = 0;
 
 
 	tas_editor_mode = TASEditorMode::DISABLED;
@@ -3128,7 +3127,6 @@ struct HwDLL::Cmd_Plus_BXT_CH_Hook
 		const auto tr = ServerDLL::GetInstance().TraceLine(view, end, 0, pl);
 
 		hw.ch_hook_point = tr.vecEndPos;
-		hw.ch_hook_hp_before = pl->v.health;
 	}
 
 	static void handler(int)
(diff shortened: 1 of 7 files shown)


khanghugo:
huh, not sure why this happens, hmm i guess i forgot to do somethings

SmileyAG:
> No more HP thing

`ch_hook_hp_before` references still presented, remark to remove all of that leftovers

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3162,11 +3162,29 @@ struct HwDLL::Cmd_Minus_BXT_CH_Hook
...
+	auto pl = hw.GetPlayerEdict();
+	auto pEngfuncs = ClientDLL::GetInstance().pEngfuncs;
+
+	auto m_iBeam = pEngfuncs->pEventAPI->EV_FindModelIndex( "sprites/smoke.spr" );
+
+	pEngfuncs->pEfxAPI->R_BeamPoints(
+		player.Origin, // start
+		ch_hook_point, // end
+		m_iBeam, // model
+		hw.GetFrameTime() * 2.0, // life
```
SmileyAG, line 3176: And there is warning about double->float conversion, so we rather have to use `2.0f`

YaLTeR:
Thanks

PR #459, opened 2023-11-15
Add bxt_ch_checkpoint

khanghugo:
This is some of the stuffs that AmxModX KZ plugins have. I am on Lenix is very hard to hook up AmxModX plugins and do things in LAN servers (which is not any easier on Windows Half-Life with BXT). So I am trying to port more things I find useful here.

This is nice because I can do check point and go check for practicing/testing things. 

I did plan to have it with un-gocheck and un-checkpoint but I think it isn't very needed because you can just noclip with cheat.

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -3202,7 +3202,7 @@ void HwDLL::ChHookPlayer() {
 	pl->v.velocity = target;
 }
 
-struct HwDLL::Cmd_BXT_CH_CheckPoint
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 {
 	NO_USAGE();
 
@@ -3238,7 +3238,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint
 	}
 };
 
-struct HwDLL::Cmd_BXT_CH_GoCheck
+struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
 {
 	NO_USAGE();
 
@@ -5363,8 +5363,8 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 		Handler<float, float, float>>("bxt_ch_set_vel_angles");
 	wrapper::AddCheat<Cmd_Plus_BXT_CH_Hook, Handler<>, Handler<int>>("+bxt_ch_hook");
 	wrapper::AddCheat<Cmd_Minus_BXT_CH_Hook, Handler<>, Handler<int>>("-bxt_ch_hook");
-	wrapper::AddCheat<Cmd_BXT_CH_CheckPoint, Handler<>>("bxt_ch_checkpoint");
-	wrapper::AddCheat<Cmd_BXT_CH_GoCheck, Handler<>>("bxt_ch_gocheck");
+	wrapper::AddCheat<Cmd_BXT_CH_CheckPoint_Create, Handler<>>("bxt_ch_checkpoint_create");
+	wrapper::AddCheat<Cmd_BXT_CH_CheckPoint_GoTo, Handler<>>("bxt_ch_checkpoint_goto");
 	wrapper::Add<
 		Cmd_BXT_Set_Angles,
 		Handler<float, float>,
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3179,6 +3181,79 @@ void HwDLL::ChHookPlayer() {
...
+
+		// not moving after go check
+		pl->v.velocity = Vector(0, 0, 0);
+
+		if (CVars::bxt_ch_checkpoint_with_vel.GetBool())
+			pl->v.velocity = hw.ch_checkpoint_vel;
+
+		pl->v.origin = hw.ch_checkpoint_origin;
+		// for CS 1.6 stamina reset
+		pl->v.fuser2 = 0;
```
SmileyAG, line 3253: I'd rather put here a bool check for cstrike directory, just for the case if that would do some the wrong stuff in some modifications

https://github.com/YaLTeR/BunnymodXT/blob/1ee1987d2b4d37ca7c0458927364a97ceee5c33e/BunnymodXT/modules/HwDLL.cpp#L6262

As you see - we doing that in code from a while ago, so just declare that variable in header

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3179,6 +3181,79 @@ void HwDLL::ChHookPlayer() {
...
+
+		auto pl = hw.GetPlayerEdict();
+
+		if (!pl)
+			return;
+
+		cl.pEngfuncs->SetViewAngles(hw.ch_checkpoint_viewangles);
+
+		if (hw.ch_checkpoint_is_duck) {
+			pl->v.flags |= 1 << 14;
```
SmileyAG, line 3241: Let's make it clear for reading, since we have included `const.h` and `in_buttons.h`

```
			pl->v.flags |= FL_DUCKING;
			pl->v.button |= IN_DUCK;
```

SmileyAG (changes requested):

SmileyAG:
I'd started to think that it might requires more attention that it was done for now

Meaning that it would be cool to store a multiple checkpoints info (angles/origin/velocity) into containers, which will allow you not to be limited by one element
Might be would be cool to add new global variable that would be in-/de-cremented if `_prev` or `_next` commands typed, that will allow to switching through multiple checkpoints by simply a pressing few times a key with binds for those

Let me write list of commands and simple comments to show how it should look in a perfect world:

```
bxt_checkpoint_create // Create checkpoint
bxt_checkpoint_reset // Reset all checkpoints
bxt_checkpoint_prev // Get to previous checkpoint (requires a variable that would be decrement each time the command insert)
bxt_checkpoint_next // Get to next checkpoint (requires a variable that would be increment each time the command insert)
bxt_checkpoint_remove // Remove last checkpoint
bxt_checkpoint_remove <element value> // Remove the specified checkpoint
bxt_checkpoint_goto <element value> // Go to the specified checkpoint
bxt_hud_checkpoint 1 // Draw number of current checkpoint and total count (example: 29/35)
```

khanghugo:
That would be very nice. My use case for now is just to do check point and that is it. If it is possible I think we should do a HUD menu like server plugins because less bind keys. The use case for it would be overkill but I can think it set up for precedence for KZ records without server plugins. And that would be way out of scope for just this simple checkpoint.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5258,6 +5341,8 @@ void HwDLL::RegisterCVarsAndCommandsIfNeeded()
 		Handler<float, float, float>>("bxt_ch_set_vel_angles");
 	wrapper::AddCheat<Cmd_Plus_BXT_CH_Hook, Handler<>, Handler<int>>("+bxt_ch_hook");
 	wrapper::AddCheat<Cmd_Minus_BXT_CH_Hook, Handler<>, Handler<int>>("-bxt_ch_hook");
+	wrapper::AddCheat<Cmd_BXT_CH_CheckPoint, Handler<>>("bxt_ch_checkpoint");
+	wrapper::AddCheat<Cmd_BXT_CH_GoCheck, Handler<>>("bxt_ch_gocheck");
```
YaLTeR, line 5345: The naming is non-obvious, you won't find the command if you don't know beforehand. I'd prefer something like `bxt_ch_checkpoint_save` and `bxt_ch_checkpoint_restore`.

YaLTeR: Or some variation on the subject. create and goto also work.

YaLTeR:
Thanks

PR #460, opened 2023-11-16
Applied fixes for bxt_fix_widescreen_fov and bxt_force_fov

### LINKED ISSUE #451: bxt_fix_widescreen_fov: breaks during the changelevel
Some explanation and fixed code:

Would close that issue with the coming soon PR

Next time I should be more conscious about testing features since I didn't even knew there was a such case due to the fact I didn't go through maps with `bxt_fix_widescreen_fov` enabled

SmileyAG:
Closes #451 

That's where journey is started! The top of first requests usually would be directed for making good *soil* to implementing features and supporting versions later, so stay in tune!


YaLTeR:
Thanks

PR #461, opened 2023-11-16
CMakeLists: added 'HLSDK10' option to define the new preprocessor flag

SmileyAG:
That's absolutely needed, since the interface versions, align of engine interfaces and size of structs is completely different in pre-WON 1.1.0.0 versions of game


YaLTeR:
a bit scared to imagine where we're heading here

PR #462, opened 2023-11-17
HLSDK10_BUILD: align of some needed structures

SmileyAG:
1) `entity_state_t`, `entvars_t` and `edict_t` is brought from [HLSDK 1.0 source code](https://github.com/a1batross/HLSDK_ancient/tree/master) (this is not a RE project, it just simply uploaded HLSDK 1.0 on GitHub)
Note: `EDICT.H` in repository is incorrect, the true `edict_t` structure in `PROGS.H` file

2) HLSDK 1.0 source code is only compatible with WON versions from 1.0.0.9 to 1.0.1.6
Half-Life: Day One, WON 1.0.0.8 or older requires the some reverse-engineering structs and interface versions to be compatible too

I used `sizeof` for those structures and compared their size to what old engine's uses in fact and I was able to confirm this thanks to the fact that I searched the engine and found some functions that actually call the size of the needed structures and it's equals

Also selectively took several variables from the structure to see what their offset was and compare it with engine references, and this also equals

Sources of where to find size of structures from those changes in engine:

## cl_entity_t

Search for `CL_EntityNum: %i is an invalid number, cl.max_edicts is %i` string to find **CL_EntityNum** function

## edict_t

Search for `NUM_FOR_EDICT: bad pointer` string to find **NUM_FOR_EDICT** function

## entvars_t

In HLSDK 2.0 version of engine, it could be found at the start of `ED_ParseEdict` function (in last argument of called **Q_memset**) 
**ED_ParseEdict** could be found from `Can't init %s` string

In older versions of engine, seems there is no references to `entvars_t` size, so, let's pretend and hope that in **edict_t** after `v` variable is nothing added

Then, we can to find the function, that calls some variable from `entvars_t` through the `edict_t` structure (example: `v.classname`)

Firstly, find the **iGetIndex** function from `globalname` string and see what offset of **classname** in that function (usually the offset is 0, unless something added at the start of **entvars_t**)
Then, find the **EntityInit** function from `Bad class!!` string and see what offset of **v.classname** in that function

Now, we doing the unga bunga math: `edict_t size - v.classname offset - classname offset = entvars_t size`
As said, that only works in case that there is not added variables after `v` in **edict_t** structure

## entity_state_t

Search for `Overflow beam entity list!` string to find **CL_LinkCustomEntity** function

### DIFF
--- a/BunnymodXT/modules/HwDLL.hpp
+++ b/BunnymodXT/modules/HwDLL.hpp
@@ -93,11 +93,11 @@ class HwDLL : public IHookableNameFilterOrdered
 		unsigned cursize;
 	};
 
-	struct svs_t
+	struct server_static_t
 	{
-		char unk[4];
+		int dll_initialized;
 		client_t *clients;
-		int num_clients;
+		int maxclients;
 	};
 
 	struct Key
@@ -216,7 +216,7 @@ public:
 		return *reinterpret_cast<double *>(reinterpret_cast<uintptr_t>(psv) + offTime);
 	}
 	inline edict_t* GetPlayerEdict() const {
-		if (!svs || svs->num_clients == 0)
+		if (!svs || svs->maxclients == 0)
 			return nullptr;
 
 		return *reinterpret_cast<edict_t**>(reinterpret_cast<uintptr_t>(svs->clients) + offEdict);
@@ -551,7 +551,7 @@ protected:
 	ptrdiff_t offNumEdicts;
 	ptrdiff_t offMaxEdicts;
 	ptrdiff_t offEdicts;
-	svs_t *svs;
+	server_static_t *svs;
 	ptrdiff_t offEdict;
 	void *svmove;
 	void **ppmove;
--- a/BunnymodXT/modules/HwDLL.hpp
+++ b/BunnymodXT/modules/HwDLL.hpp
@@ -93,12 +93,21 @@ class HwDLL : public IHookableNameFilterOrdered
 		unsigned cursize;
 	};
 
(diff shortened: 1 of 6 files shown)


SmileyAG:
Oh I forgot to explain of alignment in `server_static_t`, look here:



That's screenshot from HLDS 3.0.1.6 Linux (software engine, functions of which stripped to work for server-side only, it's last version that uses HLSDK 1.0 as base, simply: 3.0.1.6 = 1.0.1.6)

svs.0 = svs.maxclients
svs.28688 = svs.clients

clients offset (28688) - maxclients type size (4, integer) = alignment bytes

Even if there wouldn't be the HLDS Linux version, it's still possible to find by the same logic (addr2 - addr1 - addr1 type size = alignment bytes)

BunnymodXT/modules/HwDLL.hpp
```diff
@@ -93,12 +93,21 @@ class HwDLL : public IHookableNameFilterOrdered
...
+	struct server_static_t
+	{
+		int maxclients;
+		byte align[28684];
+		client_t *clients;
+	};
+	#else
 	struct server_static_t
 	{
 		int dll_initialized;
```
YaLTeR, line 106: Does BXT not use this field? Because it kinda just disappears with the define in this commit.

SmileyAG: BXT only uses `clients` and `maxclients` from that struct, so don't need to worry about

YaLTeR: I see

YaLTeR:
Thanks

PR #464, opened 2023-11-17
HLSDK10_BUILD: used the right interface version and separated from unavailable functions

SmileyAG:
`INTERFACE_VERSION` can be found in **GetEntityAPI** exported function: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/cbase.cpp#L104

`CLDLL_INTERFACE_VERSION` can be found in **Initialize** exported function: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/cl_dll/cdll_int.cpp#L151


YaLTeR:
Thanks

PR #465, opened 2023-11-17
Added bxt_allow_keypresses_in_demo

SmileyAG:
Demonstration:

https://github.com/YaLTeR/BunnymodXT/assets/58108407/192a1367-ce16-46a5-82f2-aeeaa48aa1eb

By default, when you pressing the key, engine call the console and stop doing some cook with your key inputs:

```cpp
  if ((((cls.demoplayback != false) && (cls.spectator == false)) && (down != false)) &&
     ((consolekeys[key] != false && (key_dest == key_game)))) {
    Con_ToggleConsole_f();
    return;
  }
```


YaLTeR:
Thanks

PR #468, opened 2023-11-27
Advanced checkpoints

SmileyAG:
https://github.com/YaLTeR/BunnymodXT/pull/459#issuecomment-1815195253

@khanghugo would like to know your response about that, since I really liked the idea you bring, but the implementation was so limited at the start, so I decided to modernize it

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -3322,7 +3322,7 @@ void HwDLL::ChHookPlayer() {
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 {
-	USAGE("Usage: bxt_ch_checkpoint_create\nCreates the checkpoint.\n");
+	USAGE("Usage: bxt_ch_checkpoint_create\n Creates the checkpoint.\n");
 
 	static void handler()
 	{
@@ -3434,7 +3434,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Current
 {
-	USAGE("Usage: bxt_ch_checkpoint_current\nGo to current checkpoint.\n");
+	USAGE("Usage: bxt_ch_checkpoint_current\n Go to current checkpoint.\n");
 
 	static void handler()
 	{
@@ -3446,7 +3446,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Current
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Reset
 {
-	USAGE("Usage: bxt_ch_checkpoint_reset\nReset the checkpoints.\n");
+	USAGE("Usage: bxt_ch_checkpoint_reset\n Reset the checkpoints.\n");
 
 	static void handler()
 	{
@@ -3503,7 +3503,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Remove
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Remove_After
 {
-	USAGE("Usage: bxt_ch_checkpoint_remove_after [id]\nDeletes the checkpoints following id.\n");
+	USAGE("Usage: bxt_ch_checkpoint_remove_after [id]\n Deletes the checkpoints following id.\n");
 
 	static void handler(unsigned long id)
 	{
@@ -3536,7 +3536,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Remove_After
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Next
 {
-	USAGE("Usage: bxt_ch_checkpoint_next\nGo to the next checkpoint.\n");
+	USAGE("Usage: bxt_ch_checkpoint_next\n Go to the next checkpoint.\n");
 
 	static void handler()
 	{
@@ -3563,7 +3563,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Next
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Prev
 {
-	USAGE("Usage: bxt_ch_checkpoint_prev\nGo to the previous checkpoint\n");
+	USAGE("Usage: bxt_ch_checkpoint_prev\n Go to the previous checkpoint\n");
 
 	static void handler()
 	{
@@ -3578,7 +3578,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Prev
 			}
 			else
 			{
-				hw.ORIG_Con_Printf("Not possible go to the previous checkpoint, since there are less than two of them!\n");
+				hw.ORIG_Con_Printf("Not possible go to the previous checkpoint, since the current checkpoint is the first one!\n");
 			}
 		}
 		else
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -3322,7 +3322,7 @@ void HwDLL::ChHookPlayer() {
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 {
-	USAGE("Usage: bxt_ch_checkpoint_create\n");
+	USAGE("Usage: bxt_ch_checkpoint_create\nCreates the checkpoint.\n");
 
 	static void handler()
 	{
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3270,15 +3304,135 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+{
+	USAGE("Usage: bxt_ch_checkpoint_current\n");
+
+	static void handler()
+	{
+		auto &hw = HwDLL::GetInstance();
+
+		std::ostringstream ss;
+		ss << "bxt_ch_checkpoint_goto " << hw.ch_checkpoint_current << "\n";
+		hw.ORIG_Cbuf_AddText(ss.str().c_str());
```
YaLTeR, line 3333: I'd rather call the handler directly than go through the entire commandline machinery.

YaLTeR (changes requested):

khanghugo:
Personally, I think the checkpoint previous might be the best addition here.

I am only playing KZ so that's my experience so far. When I play, I sometimes make the wrong checkpoint and that is used (very frequent). There is another case where I go check after landing the jump, so I have to "undo go check". It is nice to have. 

One quirk with using checkpoint in my case is that I will spam the checkpoint.

What I had in mind when I made this is to do map checking. Having check point helps with saving the best setup I have for some jumps to reliably do it over and over. This is where the spamming comes in. I have to stand at the very edge some times and so I have to spam check point create. For that, I think the previous check might be useful because I can check multiple jumps at the same time (obviously after getting a jump done). For normal gameplay, as I mentioned, that one more extra feature might be nice to have. The rest you have in the PR seems very overkill and personally I don't think I will find them that useful.

SmileyAG:
> Personally, I think the checkpoint previous might be the best addition here.
For normal gameplay, as I mentioned, that one more extra feature might be nice to have. The rest you have in the PR seems very overkill and personally I don't think I will find them that useful.

As the person who dedicated more than thousand hours at HLKZ, I wouldn't call `_goto`, `_remove`, `_reset` and `_next` cmds as overkill 

Because there is always practice when you want to make multiple checkpoints of same place, but with different velocity or bhop patterns in order to find out the best route for yourself

Like:
- Quickly switch between checkpoints with `_next` and `_prev` cmds
- Checkpoint that have a wrong velocity or bhop pattern could be erase with `_remove` command by specifying ID of that checkpoint
- To clear previous checkpoints and start the new series (e.g. for new map ) we could to use `_reset` command
- `_goto` could be used if I want to re-check to see if I might do better or find new route from that specific checkpoint, and if so, then I would to redo all checkpoints that go after that

I'd also started to think that it might be cool to add `bxt_ch_checkpoint_remove_after <id>`, there is use case for it:

Imagine that I have about 11 checkpoints, but everything that I done after 8th checkpoint is bad (due of missed skip, less velocity gained before, bad bhop pattern or some other reason)
So, I could type `bxt_ch_checkpoint_remove_after 8`, and it will erase all checkpoints that goes after 8th! (9, 10, 11)

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3386,6 +3386,35 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Remove
...
+				if (hw.ch_checkpoint_current > id) // If the current checkpoint has an ID greater than the specified one, then we equate it to the specified one.
+					hw.ch_checkpoint_current = id;
+
+				hw.ch_checkpoint_origin.erase(hw.ch_checkpoint_origin.begin() + id, hw.ch_checkpoint_origin.end());
+				hw.ch_checkpoint_vel.erase(hw.ch_checkpoint_vel.begin() + id, hw.ch_checkpoint_vel.end());
+				hw.ch_checkpoint_viewangles.erase(hw.ch_checkpoint_viewangles.begin() + id, hw.ch_checkpoint_viewangles.end());
+				hw.ch_checkpoint_is_duck.erase(hw.ch_checkpoint_is_duck.begin() + id, hw.ch_checkpoint_is_duck.end());
+				hw.ch_checkpoint_total = id;
+				hw.ORIG_Con_Printf("Removed the checkpoints following %lu id.\n", id);
+			}
```
YaLTeR, line 3409: Add a print in an else case too I guess saying that the id is invalid

khanghugo:
After using the normal checkpoint for a while, this starts to look appealing. I find myself playing movement modes in single player a lot and having this is very nice.

SmileyAG:
I don’t see anything else that could be changed, so this request is ready to be accepted, @YaLTeR

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3319,7 +3322,7 @@ void HwDLL::ChHookPlayer() {
 
 struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 {
-	NO_USAGE();
+	USAGE("Usage: bxt_ch_checkpoint_create\nCreates the checkpoint.\n");
```
YaLTeR, line 3325:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_create\n Creates the checkpoint.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
-		static bool is_cstrike = cl.DoesGameDirMatch("cstrike");
-		if (is_cstrike) 
+		if (hw.is_cstrike_dir) 
 			pl->v.fuser2 = 0;
 	}
 };
 
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Current
+{
+	USAGE("Usage: bxt_ch_checkpoint_current\nGo to current checkpoint.\n");
```
YaLTeR, line 3437:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_current\n Go to current checkpoint.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+	{
+		auto &hw = HwDLL::GetInstance();
+
+		HwDLL::Cmd_BXT_CH_CheckPoint_GoTo::handler(hw.ch_checkpoint_current);
+	}
+};
+
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Reset
+{
+	USAGE("Usage: bxt_ch_checkpoint_reset\nReset the checkpoints.\n");
```
YaLTeR, line 3449:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_reset\n Reset the checkpoints.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+		else
+		{
+			hw.ORIG_Con_Printf("There are no checkpoints!\n");
+		}
+	}
+};
+
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Remove_After
+{
+	USAGE("Usage: bxt_ch_checkpoint_remove_after [id]\nDeletes the checkpoints following id.\n");
```
YaLTeR, line 3506:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_remove_after [id]\n Deletes the checkpoints following id.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+		else
+		{
+			hw.ORIG_Con_Printf("There are no checkpoints!\n");
+		}
+	}
+};
+
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Next
+{
+	USAGE("Usage: bxt_ch_checkpoint_next\nGo to the next checkpoint.\n");
```
YaLTeR, line 3539:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_next\n Go to the next checkpoint.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+		else
+		{
+			hw.ORIG_Con_Printf("There are no checkpoints!\n");
+		}
+	}
+};
+
+struct HwDLL::Cmd_BXT_CH_CheckPoint_Prev
+{
+	USAGE("Usage: bxt_ch_checkpoint_prev\nGo to the previous checkpoint\n");
```
YaLTeR, line 3566:
```suggestion
	USAGE("Usage: bxt_ch_checkpoint_prev\n Go to the previous checkpoint\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3385,17 +3422,172 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_GoTo
...
+		if (!hw.ch_checkpoint_is_duck.empty())
+		{
+			if (hw.ch_checkpoint_current > 1) // Don't decrement if we have less than 2 elements
+			{
+				hw.ch_checkpoint_current--;
+				HwDLL::Cmd_BXT_CH_CheckPoint_GoTo::handler(hw.ch_checkpoint_current);
+			}
+			else
+			{
+				hw.ORIG_Con_Printf("Not possible go to the previous checkpoint, since there are less than two of them!\n");
```
YaLTeR, line 3581:
```suggestion
				hw.ORIG_Con_Printf("Not possible go to the previous checkpoint, since the current checkpoint is the first one!\n");
```

YaLTeR (approved):

YaLTeR:
Thanks!

PR #469, opened 2023-11-28
Added bxt_get_steamid_from_demo

SmileyAG:
I know that there is `uint64_t`, but I decided to took from HLSDK to keep the same behaviour (for case if `uint64_t` is differs on some other platform and it will go for the fancy-schmancy stuff): https://github.com/ValveSoftware/halflife/blob/de70fada5061965884492178a7d6166646424a4e/public/steam/steamtypes.h

That `76561197960265728` const is not something that I got fantasied out of mind, most of Steam-related and demo parsing tools in fact uses the same *magic* number that is actually the starting ID

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -1397,9 +1397,11 @@ void HwDLL::FindStuff()
 				default:
 				case 0: // HL-Steampipe
 					ClientDLL::GetInstance().pEngfuncs = *reinterpret_cast<cl_enginefunc_t**>(reinterpret_cast<uintptr_t>(ClientDLL_Init) + 181);
+					is_steamid_build = true;
 					break;
 				case 1: // HL-4554
 					ClientDLL::GetInstance().pEngfuncs = *reinterpret_cast<cl_enginefunc_t**>(reinterpret_cast<uintptr_t>(ClientDLL_Init) + 226);
+					is_steamid_build = true;
 					break;
 				case 2: // HL-NGHL
 					ClientDLL::GetInstance().pEngfuncs = *reinterpret_cast<cl_enginefunc_t**>(reinterpret_cast<uintptr_t>(ClientDLL_Init) + 203);
@@ -1409,6 +1411,7 @@ void HwDLL::FindStuff()
 					break;
 				case 4: // CoF-5936
 					ClientDLL::GetInstance().pEngfuncs = *reinterpret_cast<cl_enginefunc_t**>(reinterpret_cast<uintptr_t>(ClientDLL_Init) + 230);
+					is_steamid_build = true;
 					break;
 				}
 			});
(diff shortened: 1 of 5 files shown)


HLSDK/common/com_model.h
```diff
@@ -420,6 +420,10 @@ typedef struct player_info_s
 	vec3_t	prevgaitorigin;
 
 	customization_t customdata;
+
+	char	hashedcdkey[16];
+
+	uint64	m_nSteamID;
```
YaLTeR, line 426: Are these fields only available on the Steam engines? Would be good to add a comment.

BunnymodXT/modules/HwDLL.hpp
```diff
@@ -219,6 +219,7 @@ class HwDLL : public IHookableNameFilterOrdered
 	inline bool IsActive() { return (psv && *reinterpret_cast<int*>(psv)); }
 
 	inline bool IsRecordingDemo() const { return demorecording && *demorecording == 1; }
+	inline bool IsPlayingbackDemo() const { return demoplayback && *demoplayback == 1; }
```
YaLTeR, line 222:
```suggestion
	inline bool IsPlayingDemo() const { return demoplayback && *demoplayback == 1; }
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3486,6 +3489,32 @@ struct HwDLL::Cmd_BXT_Get_Server_Time
...
+
+			const uint64 STEAMID64_CONST = 76561197960265728; // 0x110000100000000
+			const unsigned long STEAMID32 = static_cast<unsigned long>(player_info->m_nSteamID);
+			const uint64 STEAMID32_TO_64 = STEAMID64_CONST + STEAMID32;
+
+			hw.ORIG_Con_Printf("SteamID32: %" PRIu64 "\n", STEAMID32);
+
+			std::ostringstream ss;
+			ss << "SteamID64: " << STEAMID32_TO_64 << "\n";
+			hw.ORIG_Con_Printf(ss.str().c_str());
```
YaLTeR, line 3513:
```suggestion
			hw.ORIG_Con_Printf("SteamID64: %" PRIu64 "\n", STEAMID32_TO_64);
```

YaLTeR (changes requested):
> I know that there is uint64_t, but I decided to took from HLSDK to keep the same behaviour (for case if uint64_t is differs on some other platform and it will go for the fancy-schmancy stuff)

Could you please just use `uint64_t`? The entire point of a type like this that it is the exact same on all platforms. If it turns out that at Valve this is not the case then call the alias something else because at that point it most certainly does not qualify to be called "uint64".

HLSDK/common/com_model.h
SmileyAG: Can confirm that starting from 4554 build - `m_nSteamID` field exists.

BunnymodXT/modules/HwDLL.hpp
SmileyAG: https://github.com/YaLTeR/BunnymodXT/blob/89595432c52fcf01f5339f94b56cab1a986bbf88/HLSDK/common/demo_api.h#L23-L24

BunnymodXT/modules/HwDLL.cpp
SmileyAG: Let me tell you a little backstory behind that: yes, I tried to pass it directly but the output value was a totally different (the format and types is still being the same)

And in one moment I got the idea in mind: *what if I simply pass that value as string instead*? And... that really started to output a right value!

Casting `STEAMID32` from **unsigned long long** to **unsigned long** also helped to solve that issue

SmileyAG:
> then call the alias something else because at that point it most certainly does not qualify to be called "uint64".

What you'd like to prefer from next ones? `steam_uint64`, `valve_uint64`, `uint64_steam`, `uint64_valve`, `S_uint64` or `V_uint64`? (the last two is got from idea that Quake developers re-implemented standard functions as `strcmp` at their own and named it starting with `Q_` prefix and GoldSrc code still refers to them in quite a half of functions)

Jokes aside, I could also accept for changing to `uint64_t` since that really not a big deal for implement at our own, so I leave the decision up to you

BunnymodXT/modules/HwDLL.cpp
YaLTeR: That is quite strange.

YaLTeR:
> What you'd like to prefer from next ones?

Could you maybe just call it `steamid_t` or something

YaLTeR:
Thanks

PR #470, opened 2023-11-28
Added bxt_cof_allow_to_skip_all_cutscenes

SmileyAG:
Quickly saying, in CoF it's possible to skip majority of cutscenes by simply holding the `ENTER` key (game is checking for `ENTER` key code in Windows API, and if so, it will speed up the game by set **host_framerate**)

But, there a only a like at least one cutscene that cannot be skippable (due of that mapper set a spawn flag for them), so runners does the set `host_framerate` manually instead! (funny and pity situation, when the game stop doing that - people start doing that themself)

So in the future, when leaderboard would switched to mandatory with BXT and rules adapted for that, it could be simply erase from that bad habit by enabling `bxt_cof_allow_skipping_all_cutscenes 1`

https://github.com/LogicAndTrick/sledge-formats/blob/05b3fcfb6261b956120e4d218e7a5a357e42ac51/Sledge.Formats.GameData.Tests/Resources/fgd/goldsource/cry-of-fear.fgd#L3720

### DIFF
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -90,7 +90,7 @@
 	X(bxt_cof_slowdown_if_use_on_ground, "0") \
 	X(bxt_cof_disable_monsters_teleport_to_spawn_after_load, "0") \
 	X(bxt_cof_disable_viewpunch_from_jump, "0") \
-	X(bxt_cof_allow_to_skip_all_cutscenes, "0") \
+	X(bxt_cof_allow_skipping_all_cutscenes, "0") \
 	X(cl_righthand, "0") \
 	X(bxt_remove_punchangles, "0") \
 	X(bxt_disable_brush_entities, "0") \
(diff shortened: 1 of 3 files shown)


BunnymodXT/cvars.hpp
```diff
@@ -90,6 +90,7 @@
 	X(bxt_cof_slowdown_if_use_on_ground, "0") \
 	X(bxt_cof_disable_monsters_teleport_to_spawn_after_load, "0") \
 	X(bxt_cof_disable_viewpunch_from_jump, "0") \
+	X(bxt_cof_allow_to_skip_all_cutscenes, "0") \
```
YaLTeR, line 93:
```suggestion
	X(bxt_cof_allow_skipping_all_cutscenes, "0") \
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2968,6 +2982,33 @@ HOOK_DEF_3(ServerDLL, void, __fastcall, CChangeLevel__TouchChangeLevel, void*, t
 	return ORIG_CChangeLevel__TouchChangeLevel(thisptr, edx, pOther);
 }
 
+HOOK_DEF_1(ServerDLL, void, __fastcall, CTriggerCamera__FollowTarget, void*, thisptr)
+{
+	entvars_t *pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(thisptr) + 4);
+	if (pev)
+	{
+		bool ret = false;
```
YaLTeR, line 2990: `ret` is one of the names I use to indicate the function return value, so mind changing this to something else? Like `changed`

YaLTeR:
Thanks

PR #471, opened 2023-11-30
Pattern-less implementation of GiveNamedItem function from HLSDK

SmileyAG:
That planned for partially support of the co-op, so that you can give the needed weapons to players without having to hook `GiveNamedItem` for the each specific mod

Code from HLSDK: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/player.cpp#L3322

If you want to test, then for example you can add the following code to the **ClientCommand** function and then type `bxt_give weapon_shotgun` in console

```cpp
} else if ((std::strcmp(cmd, "bxt_give") == 0) && CVars::sv_cheats.GetBool()) {
		int iszItem = pEngfuncs->pfnAllocString(pEngfuncs->pfnCmd_Argv(1)); // Make a copy of the classname
		GiveNamedItem(pev, iszItem);
```

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -269,6 +269,8 @@ void ServerDLL::Clear()
 	ORIG_CTriggerCamera__FollowTarget = nullptr;
 	ORIG_PM_CheckStuck = nullptr;
 	ORIG_CBaseEntity__FireBullets = nullptr;
+	ORIG_DispatchSpawn = nullptr;
+	ORIG_DispatchTouch = nullptr;
 	ORIG_CBaseEntity__FireBullets_Linux = nullptr;
 	ORIG_CBaseEntity__FireBulletsPlayer = nullptr;
 	ORIG_CBaseEntity__FireBulletsPlayer_Linux = nullptr;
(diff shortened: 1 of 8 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2616,6 +2626,28 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, ClientCommand, edict_t*, pEntity)
...
+void ServerDLL::GiveNamedItem(entvars_t *pev, int istr)
+{
+	auto &hw = HwDLL::GetInstance();
+
+	if (pEngfuncs && hw.ppGlobals && ORIG_DispatchSpawn && ORIG_DispatchTouch)
+	{
+		edict_t *pent = pEngfuncs->pfnCreateNamedEntity(istr);
+		if (!pent || !pEngfuncs->pfnEntOffsetOfPEntity(pent))
+		{
+			#define ALERT(at, format, ...) pEngfuncs->pfnAlertMessage(at, const_cast<char*>(format), ##__VA_ARGS__)
```
YaLTeR, line 2638: Just write it inline, no point in defining for 1 line

YaLTeR (changes requested):

BunnymodXT/modules/ServerDLL.cpp
chinese-soup: https://github.com/YaLTeR/BunnymodXT/blame/d0d183a4f29d1f6f1c8a413dae132bca586df978/BunnymodXT/modules/ServerDLL.cpp#L1919

:-)

YaLTeR: image

image

chinese-soup: Cool my passive aggressive man, but my point was that this code is also in ServerDLL.cpp, so it can be defined at the top of ServerDLL.cpp?

YaLTeR: Yeah it could

SmileyAG:
I slightly changed the concept of this request. `bxt_give` command has been added that uses this our custom code, so that it is now possible to give weapons without searching for patterns for each mod, so all **CBasePlayer::GiveNamedItem** hooks have been removed and therefore `give` is no longer handled in the code either

I personally have nothing more to contribute to this request, so it is ready to be accepted, @YaLTeR

YaLTeR (approved):

YaLTeR:
Thanks!

PR #472, opened 2023-11-30
Added bxt_fix_changelevel_in_coop

SmileyAG:
Allows the game to changelevel while co-op is enabled for all modifications (since that change done for engine-side only)

The check for `CVars::coop.GetBool()` can be replaced with `ppGlobals->coop`
But I don’t think that this makes any sense, because in fact, `ppGlobals->coop` simply writes the value of `coop` cvar


YaLTeR:
What's the use case?

SmileyAG:
> What's the use case?

Isn't that clearly said in description of request? Well, simply to say, if the game want to make changelevel with specified landmark name - we make it empty instead, so the engine would call `changelevel` instead of `changelevel2`, that is compatible with multiplayer

Since you know that there are unique games like Gunman Chronicles, for which the source code is not available, but you still want to have at least some grains of co-op

Of course, you can take .bsp and remove name in **landmark** argument for the `trigger_changelevel` entity for all maps to get the same output so as not to use third-party hook tools like BXT, but are you really ready to do this for every mod?

For the same reason, I would like for example, save the position of the player who activated the changelevel, and then after adjusting it with the landmark offset for next map, make it set to all other players who will spawn next
(but then would be possibility of getting stuck in each other, then maybe give them a non-solid state for a while, or in the worst case, make an offset to a some units when spawning, so that they spawn a little to the left or to the right, but that can end up into stuck in wall or clipping through some wall, so it requires some checks for that case too)
Because other people who want to do this without any code changes would simply move `info_player_start` entity for all maps manually

BXT has long been positioned as a Swiss knife, in other words, what may seem like a stone to one person may seem like a diamond to another (for one may think that adding a new variable is useless and it is enough for him to have a hardcoded value, while for another, having a separate variable will be useful)

SmileyAG:
> BXT has long been positioned as a Swiss knife, in other words, what may seem like a stone to one person may seem like a diamond to another (for one may think that adding a new variable is useless and it is enough for him to have a hardcoded value, while for another, having a separate variable will be useful)

I forgot to mention that I'm not planning to fullfill the BXT with huge scale of co-op implementation, I can understand it may too much for such a little thing, but having some basic grains of that would be cool anyway

YaLTeR:
> Isn't that clearly said in description of request?

The description says what this does, not why. I wasn't aware that there even is coop in GoldSrc (isn't that a Decay thing?).

> BXT has long been positioned as a Swiss knife

Not really? The summary is "Speedrun and TAS tool for Half-Life & friends." and nothing in the README says otherwise. I understand that over the past years it had grown considerably, but still most of those additions were mainly something that could be useful for various kinds of speedruns.

I'm not saying here that I'm against this specific PR (especially since it's small and not reaching too wide across the codebase). However, you need to consider that the latest version of BXT is a requirement for speedruns, and the more functionality is added, the more chance there is for some bugs to find their way in, that may affect speedrunners, even when they do not use that functionality. That is to say, if something is relatively self-contained and doesn't use many hooks, it's probably fine, but something big and not directly related to speedrunning might need a stronger justification for inclusion.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -7010,6 +7029,15 @@ HOOK_DEF_0(HwDLL, void, __cdecl, Host_Changelevel2_f)
 	insideHost_Changelevel2_f = false;
 }
 
+HOOK_DEF_2(HwDLL, void, __cdecl, PF_changelevel_I, char*, s1, char*, s2)
+{
+	// If landmark name is empty (s2), then the function will call "changelevel" instead of "changelevel2"
+	if (CVars::bxt_fix_changelevel_in_coop.GetBool() && CVars::coop.GetBool())
+		ORIG_PF_changelevel_I(s1, 0);
```
YaLTeR, line 7036:
```suggestion
		ORIG_PF_changelevel_I(s1, NULL);
```

SmileyAG:
> That is to say, if something is relatively self-contained and doesn't use many hooks, it's probably fine, but something big and not directly related to speedrunning might need a stronger justification for inclusion.

That's a fair point, but if you want to know, I have rethought my vision of some things a little over the past months, and I will say that I am trying to get rid of unnecessary hooks and patterns

I don’t have a goal that I want to fill BXT with everything just because why not, this is not true at all, each of the things that I submitted was useful to me and other people to one degree or another

In addition, I always check the validity of the patterns for the main games and some basic things before the release occurs to make sure that nothing got break

I'm going to eventually remove patterns for really unclaimed and unnecessary mods, and leave them for really ones that really worth or those that have activity

Because my earlier one added extra patterns for some mods that absolutely no one plays, and I understand now that this is a mistake, because it simply creates destructibility in theory, if we say this pattern is short and it intersects in the main games

And even if someone wants to prove that support for the X mod needs to be returned, well, let them create an issue and explain it as it is

YaLTeR:
Thanks

PR #473, opened 2023-11-30
feature: command to disable checking for cheat cmds when viewing the demo

SmileyAG:


YaLTeR:
So what does this do exactly?

SmileyAG:
> So what does this do exactly?

When playbacking demos that recorded in multiplayer, protected cvars would give the `Can't set %s in multiplayer` console message
To fix that, we have to fake our `cls.state` to pass through the **Cvar_Command** function and after we can restore it to original state

```cpp
          if ((((*(byte *)&pcVar4->flags & 0x40) != 0) && (ca_disconnected < cls.state)) &&
             (1 < cl.maxclients)) {
            Con_Printf("Can\'t set %s in multiplayer\n",pcVar4->name);
            return 1;
          }
```

And, there is `R_ForceCvars` function which calling only if `maxplayers > 1` and that forcing values to default for specific list of cvars:

```
r_lightmap
gl_clear
r_novis
r_fullbright
snd_show
chase_active
lambert
gl_monolights
gl_wireframe
r_dynamic
gl_alphamin
gl_max_size
gl_polyoffset
r_drawentities
lightgamma
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -7770,6 +7834,74 @@ HOOK_DEF_1(HwDLL, qboolean, __cdecl, CL_CheckGameDirectory, char*, gamedir)
...
+	int *state;
+	int orig_state;
+	bool return_orig_value = false;
+
+	// We can avoid the multiplayer check if cls.state <= 1
+	if (cls && CVars::bxt_disable_cheats_check_in_demo.GetBool() && IsPlayingbackDemo())
+	{
+		state = reinterpret_cast<int*>(cls);
+		orig_state = *state;
+		*state = 1;
```
YaLTeR, line 7848: cls.state is kind of an important variable. If you set it to 1, do commands like disconnect, changelevel, etc. still work?

SmileyAG: That's a fair point. It would be more rationally to replace it with set `cl.maxclients` to **1** instead.

I just didn’t want to make offsets for find that new variable through other functions before, but now I looked at the `Cvar_Command` code in more detail, and it actually calls a function for which the **cls.state** values will affect its code (`Cvar_DirectSet`)

In the case of `cl.maxplayers` I can't tell at first glance, but even if it may turn out to be the case later, it will be less harmful than changing `cls.state`

PR #477, opened 2023-12-22
HwDLL: Warn users about encrypted client dll

hobokenn:
This happens fairly often when double-clicking hl.exe and injecting manually on older engines. The hook is added as fail-safe if someone deliberately tries launching with injector in the wrong gamedir.

edit: accidentally force-pushed without committing, which automatically closed previous PR.


YaLTeR:
Thanks

PR #480, opened 2023-12-30
Inherit msurface_t for new structure due to added variable in the end with Half-Life 25th anniversary update

SmileyAG:
You can't just add it to the end of the current structure, because this will simply break compatibility with all previous versions of the engine when using data from `msurface_t` for BXT code

All you need is to say create a variable in the code, like: `bool is_hl25_update = false` and enable it if BXT detects that we definitely are playing on a new version of the engine

After it should be possible to use `reinterpret_cast` in places where that related code is used to keep compatibility for both versions

### DIFF
--- a/HLSDK/common/com_model.h
+++ b/HLSDK/common/com_model.h
@@ -272,6 +272,19 @@ struct msurface_s
 };
 #endif
 
+typedef struct displaylist_s // Half-Life 25th anniversary update (hardware engine)
+{
+	unsigned gl_displaylist;
+	int rendermode;
+	float scrolloffset;
+	int renderDetailTexture;
+} displaylist_t;
+
+struct msurface_hw_25th_anniversary_t : public msurface_t 
+{
+	displaylist_t displaylist; // Half-Life 25th anniversary update (hardware engine)
+};
+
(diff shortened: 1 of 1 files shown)


YaLTeR:
So the C way of doing it is

```c
typedef struct {
    msurface_t surface;
    displaylist_t displaylist;
} msurface_hw_25th_anniversary_t;
```

I'd prefer to do that instead.

SmileyAG:
> So the C way of doing it is
> 
> ```c
> typedef struct {
>     msurface_t surface;
>     displaylist_t displaylist;
> } msurface_hw_25th_anniversary_t;
> ```
> 
> I'd prefer to do that instead.

I really don't like that approach, because imagine when for one case it would be `var->not_inherited.x` and for another it would be `var->x`

It would be better if both of them use `var->x`, but simply casted to according struct for that game version

YaLTeR:
Mh. Well, alright. Just that there's a million rules in C++ for when the derived class like this is laid out in memory as you expect. If we do something by accident that breaks those rules and causes the layout to change, we won't even notice until someone actually tries to use it in-game. The C way does not have this problem.

PR #481, opened 2023-12-30
Removed GPL copyright from Quake for cmd.h

SmileyAG:
🤫

But seriously speaking, the comments remaining in this header are also appear in the Source SDK code, so I think it fair to remove only the copyright

After all, in fact, this structure was is not full copy but a bit reversed from GoldSrc, example that `int flags;` in the end of **cmd_function_t** is not exist in header from Quake code, I got it added from decompiled linux goldsrc engine code


YaLTeR:
Yeah besides I'm not sure how copyrightable this small header would be

PR #500, opened 2024-02-08
Add bxt_enable_big_map

### LINKED ISSUE #499: Support of loading larger maps
Huge inspiration after seeing maps bigger than default +-4k size.

There needs to be changes to delta.lst and server DLL function `IsInWorld()`. 

BXT can easily change these on the fly without modifying the game files. For delta.lst, override certain value in certain static variable after `CL_Parse_DeltaDescription()` when loading the map. `IsInWorld()` cannot be patched from what I see in Ghidra. However, the function is very simple to replicate and we can ignore its result in favor of what we want (bigger number)

khanghugo:
Closes #499 

Here is an apparent big map.

[arte_marte (2).zip](https://github.com/YaLTeR/BunnymodXT/files/14203119/arte_marte.2.zip)

~(waiting for CI to test Windows build)~


SmileyAG:
Requested changes in https://github.com/khanghugo/BunnymodXT/pull/2

SmileyAG:
And also, can you make this also optionally disabled?

If not, then I think it will be possible to make custom bit flags and save them in BXT data for specific feature enabled, so you will know if the runner forgot to restart the game to disabling this option, that would be fine solution too

khanghugo:
What do you mean by optionally disabled? The feature as a whole or the big map when I don't want to have big map anymore? The former I guess could be done if they have a cvar enable to have them enable this? LOL. 

As for the latter, it is just a lot more complicated. Like not complicated but cumbersome. Deep copy the struct then restore it back, something on the top of my head. I just guess that people running this command will just do stuffs in the big map and don't do anything else. 

The flag thing seems like a good choice though that is not really my domain and I definitely won't know how to test it.

SmileyAG:
> though that is not really my domain and I definitely won't know how to test it.

Requested in https://github.com/khanghugo/BunnymodXT/pull/3

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1566,6 +1579,19 @@ void ServerDLL::FindStuff()
...
+	{
+		auto pattern = fCBaseEntity__IsInWorld.get();
+		if (ORIG_CBaseEntity__IsInWorld) {
+			EngineDevMsg("[server dll] Found CBaseEntity::IsInWorld at %p (using the %s pattern).\n", ORIG_CBaseEntity__IsInWorld, pattern->name());
+		} else {
+			ORIG_CBaseEntity__IsInWorld_Linux = reinterpret_cast<_CBaseEntity__IsInWorld_Linux>(MemUtils::GetSymbolAddress(m_Handle, "_ZN11CBaseEntity9IsInWorldEv"));
+			if (ORIG_CBaseEntity__IsInWorld_Linux)
+				EngineDevMsg("[server dll] Found CBaseEntity::IsInWorld [Linux] at %p.\n", ORIG_CBaseEntity__IsInWorld_Linux);
+			else
+				EngineDevWarning("[server dll] Could not find CBaseEntity::IsInWorlds.\n");
```
YaLTeR, line 1591:
```suggestion
				EngineDevWarning("[server dll] Could not find CBaseEntity::IsInWorld.\n");
```

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5205,6 +5238,43 @@ struct HwDLL::Cmd_BXT_Skybox_Reload
...
+Usage: bxt_enable_big_map\n\n\
+After entering this command in main menu, you can load maps beyond +-4096 limit.\n\
+Due to shortcomings of the implementation, you must restart your game in order to revert the effect.\n\
+Can be called in command line option when start up the game.\n");
+
+	static void handler()
+	{
+		auto &hw = HwDLL::GetInstance();
+
+		if (hw.g_sv_delta == NULL)
```
YaLTeR, line 5267: Should this also check for IsInWorld?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -5205,6 +5238,43 @@ struct HwDLL::Cmd_BXT_Skybox_Reload
...
+	}
+}
+
+struct HwDLL::Cmd_BXT_Enable_Big_Map
+{
+	USAGE("\
+Usage: bxt_enable_big_map\n\n\
+After entering this command in main menu, you can load maps beyond +-4096 limit.\n\
+Due to shortcomings of the implementation, you must restart your game in order to revert the effect.\n\
+Can be called in command line option when start up the game.\n");
```
YaLTeR, line 5261:
```suggestion
Can be called as a command line argument when starting up the game.\n");
```

BunnymodXT/modules/HwDLL.cpp
khanghugo: Would that make it not work until a server is started because the pointer is yet found? This big map works by changing deltas before starting the server. Delta (delta.lst) is loaded upon starting the game, not the server.

The `IsInWorld` function is auxiliary compared to delta change, it is more for a select few entities to behave in big map.

YaLTeR: I see

YaLTeR:
Thanks

PR #502, opened 2024-02-09
Add bxt_ch_trigger_tp_keeps_momentum

khanghugo:
This is useful for some surf/bhop maps that take advantage of portal-like teleportation.

Implementation is simple because in order for it to work, the mapper has to make the map linear as well.

This case is promised to never happen.
image

~Waiting for CI to test on Windows.
Currently crashing on Windows.....~

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -108,6 +108,11 @@ extern "C" int __cdecl _ZN11CBaseEntity9IsInWorldEv(void *thisptr)
 {
 	return ServerDLL::HOOKED_CBaseEntity__IsInWorld_Linux(thisptr);
 }
+
+extern "C" void __cdecl _ZN12CBaseTrigger13TeleportTouchEP11CBaseEntity(void* thisptr, void* pOther)
+{
+	return ServerDLL::HOOKED_CBaseTrigger__TeleportTouch_Linux(thisptr, pOther);
+}
 #endif
 
 void ServerDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* moduleBase, size_t moduleLength, bool needToIntercept)
@@ -172,7 +177,9 @@ void ServerDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* m
 			ORIG_PM_Duck, HOOKED_PM_Duck,
 			ORIG_PM_UnDuck, HOOKED_PM_UnDuck,
 			ORIG_CBaseEntity__IsInWorld, HOOKED_CBaseEntity__IsInWorld,
-			ORIG_CBaseEntity__IsInWorld_Linux, HOOKED_CBaseEntity__IsInWorld_Linux);
+			ORIG_CBaseEntity__IsInWorld_Linux, HOOKED_CBaseEntity__IsInWorld_Linux,
+			ORIG_CBaseTrigger__TeleportTouch, HOOKED_CBaseTrigger__TeleportTouch,
+			ORIG_CBaseTrigger__TeleportTouch_Linux, HOOKED_CBaseTrigger__TeleportTouch_Linux);
 	}
 }
 
@@ -227,7 +234,9 @@ void ServerDLL::Unhook()
 			ORIG_PM_Duck,
 			ORIG_PM_UnDuck,
 			ORIG_CBaseEntity__IsInWorld,
-			ORIG_CBaseEntity__IsInWorld_Linux);
+			ORIG_CBaseEntity__IsInWorld_Linux,
+			ORIG_CBaseTrigger__TeleportTouch,
+			ORIG_CBaseTrigger__TeleportTouch_Linux);
 	}
 
 	Clear();
@@ -301,6 +310,8 @@ void ServerDLL::Clear()
 	ORIG_PM_UnDuck = nullptr;
 	ORIG_CBaseEntity__IsInWorld = nullptr;
 	ORIG_CBaseEntity__IsInWorld_Linux = nullptr;
+	ORIG_CBaseTrigger__TeleportTouch = nullptr;
+	ORIG_CBaseTrigger__TeleportTouch_Linux = nullptr;
 	ppmove = nullptr;
 	offPlayerIndex = 0;
 	offOldbuttons = 0;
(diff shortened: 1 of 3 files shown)


SmileyAG:
Do you want to know what the main issue with the that implementation? The fact is that your speed needs to be redirected depending on your direction

The angles must be taken either from those specified in `info_teleport_destination` (the mapper must take care of this so that it specify the correct angles) or redirected from the player angles at the worst

Take a look to that implementation and description: https://github.com/fireblizzard/agmod/pull/11

khanghugo:
Right, then there should be two options for it. With redirect and without it. Redirect is nice but I mean to have this for big surf maps and they are guaranteed to be built like that.

This is an example of a map that only works without redirect.

https://www.youtube.com/watch?v=s6Y5BY6-iGY

I can add both of that.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3320,4 +3347,53 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		pev->v_angle = prev_view;
+	}
+}
+
+typedef int (CBaseEntity__IsPlayer_Linux)(void);
+
+HOOK_DEF_2(ServerDLL, void, __cdecl, CBaseTrigger__TeleportTouch_Linux, void*, thisptr, void*, pOther)
+{
+	auto vtable = *reinterpret_cast<uintptr_t *>(pOther);
+	auto pf = reinterpret_cast<CBaseEntity__IsPlayer_Linux*>(*reinterpret_cast<uintptr_t **>(vtable + 0xa0)); // 0x28 * 4
```
SmileyAG, line 3379: It is not efficient to hooking that type of function for each specific mod, since it can be implemented at own side:

```cpp
entvars_t* pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(thisptr) + 4);
if (IsPlayer(pev->pContainingEntity))
```

```cpp
bool ServerDLL::IsPlayer(edict_t *ent)
{
	// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/player.cpp#L2850

	auto &hw = HwDLL::GetInstance();

	if (strcmp(hw.GetString(ent->v.classname), "player") != 0)
		return false;

	if (!(ent->v.flags & FL_CLIENT))
		return false;

	if (pEngfuncs && hw.ppGlobals)
	{
		int index = pEngfuncs->pfnIndexOfEdict(ent);

		if ((index < 1) || (index > hw.ppGlobals->maxClients)) // gGlobalVariables.maxClients = svs.maxclients
			return false;
	}

	return true;
}
```

SmileyAG (changes requested):

khanghugo:
Thanks for the help. This should be good. I don't have redirect option on by default because I think maps that take advantage of this are linear.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3320,4 +3350,109 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+
+HOOK_DEF_3(ServerDLL, void, __fastcall, CBaseTrigger__TeleportTouch, void*, thisptr, int, edx, void*, pOther)
+{
+	auto is_bxt_ch_trigger_tp_keeps_momentum_enabled = CVars::sv_cheats.GetBool() && CVars::bxt_ch_trigger_tp_keeps_momentum.GetBool();
+
+	entvars_t *pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(pOther) + 4);
+	Vector prev_vel;
+	Vector prev_view;
+
+	if (pev) {
```
SmileyAG, line 3386: I suppose you should save and restore `pev->angles` and `pev->basevelocity`, since both of them is also reset by that function

SmileyAG (changes requested):

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3320,4 +3350,123 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+				pev->basevelocity.x = vecForward.x * xy_basevel;
+				pev->basevelocity.y = vecForward.y * xy_basevel;
+			} else {
+				pev->velocity = prev_vel;
+				pev->basevelocity = prev_basevelocity;
+			}
+		}
+
+		if (CVars::bxt_ch_trigger_tp_keeps_momentum_viewangles.GetBool()) {
+			pev->fixangle = 0; // cannot change angle if it is 1
```
YaLTeR, line 3426: When is this 1? Maybe we don't need to change angle if it's 1?

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3320,4 +3350,123 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+{
+	auto is_bxt_ch_trigger_tp_keeps_momentum_enabled = CVars::sv_cheats.GetBool() && CVars::bxt_ch_trigger_tp_keeps_momentum.GetBool();
+
+	entvars_t *pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(pOther) + 4);
+	Vector prev_vel;
+	Vector prev_view;
+
+	if (pev) {
+		prev_vel = pev->velocity;
+		prev_view = pev->v_angle;
```
YaLTeR, line 3443: Why does this do less things than the Windows hook?

Also it's complex enough; maybe worth extracting to a common function.

BunnymodXT/modules/ServerDLL.cpp
khanghugo: I think something going on with all of the inheritance thing so it is hard for me to track but if `pev->fixangle = 1` then when I change the angles with following lines they don't work. If I set it back to `1` after changing the angle, doesn't work either. So it has to be `0` throughout that process.

BunnymodXT/modules/ServerDLL.cpp
khanghugo: i forgo :skull:

BunnymodXT/modules/ServerDLL.cpp
khanghugo: Hmm, so the `fixangle` is handled by the engine, not the game code. In `SV_RunCmd`, it won't update the player angle if that is the case and will set it back to 0 afterward.

YaLTeR: So, when is this 1 in-game? Maybe when it is 1 we should not change the angle, instead of forcing it to 0 and changing the angle?

khanghugo: IIRC in teleport touch, hlsdk uses something else to change player viewangles and set this to 1. It is only set to 1 for teleport touch.

YaLTeR: Oh, I see

YaLTeR: Can you add that to the comment in the code please?

YaLTeR:
Thanks

PR #504, opened 2024-02-15
Add bxt_ch_fix_sticky_slide

khanghugo:
Some thing like this https://github.com/ValveSoftware/halflife/issues/3586. 

You can get the big map mention in the big map PR #500, go to this coordinate `-222.814835 -18311.080078 29012.031250`, and try to slide down all the way down.

I am testing some big maps and getting stuck on geometry is very sad. Most servers have this fix.

~~This should not be used in TAS~~ Maybe I will use for TAS, this works so well. This is only for normal gameplay.

Attribution: This is almost a one to one implementation of an AmxModX plugin from Lopol2010.

### DIFF
--- a/BunnymodXT/cvars.hpp
+++ b/BunnymodXT/cvars.hpp
@@ -234,7 +234,9 @@
 	X(bxt_splits_end_on_last_split, "0") \
 	X(bxt_ch_hook_speed, "869") \
 	X(bxt_ch_checkpoint_with_vel, "1") \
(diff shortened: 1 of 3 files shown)


khanghugo:
Hmmm, I just realize that it is possible for the slide to be from the top as well and this would just make them extra stuck. This PR is mostly concerned and about slides, not wall. I will add some kind of player trace to check where to displace the player correct er

YaLTeR:
Thanks

PR #506, opened 2024-02-21
bxt_enable_big_map: also big velocity

khanghugo:
Someone told me that they lag when going too fast. Current delta only supports +-4096. I make it +-8192. I think anything bigger than that is very impractical.

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -3276,10 +3276,8 @@ HOOK_DEF_1(ServerDLL, void, __fastcall, CBasePlayer__Jump, void*, thisptr)
 
 int ServerDLL::IsInWorld(Vector origin, Vector velocity, int map_size) // https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/cbase.cpp#L706
 {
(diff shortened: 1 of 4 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3276,10 +3276,8 @@ HOOK_DEF_1(ServerDLL, void, __fastcall, CBasePlayer__Jump, void*, thisptr)
 
 int ServerDLL::IsInWorld(Vector origin, Vector velocity, int map_size) // https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/cbase.cpp#L706
 {
-	// Copy pasted from HLSDK, but origin value is changed
-	// Maybe in the future we should also make velocity check to use sv_maxvelocity value,
-	// but I don't see any side effect when going beyond it ever
-
+	auto max_velocity = CVars::sv_maxvelocity.GetFloat();
```
SmileyAG, line 3279: Can you do this in exactly the same way as with `map_size`?

That mean to specify this when calling a function in the arguments, it gives more flexibility of use for different code

SmileyAG (changes requested):

YaLTeR (approved):

YaLTeR:
Thanks

PR #507, opened 2024-02-21
Add bxt_ch_noclip_speed

### LINKED ISSUE #505: Setting noclip velocity
Hook works great. But then when I try to navigate these tinily huge maps, I bump into teleport all the time.

This has been something on my mind for a while. 

My idea for the implementation is changing `pmove->maxspeed` and `pmove->clientmaxspeed` before `PM_CheckParameters()` during noclip and restoring it back when not in noclip. Because it is a cheat command, I don't think it would interfere with anything speedrunning.

Additionally, there could be something extra like "+speed" to double the noclip velocity.

khanghugo:
Closes #505 

By default noclip is now faster. I think it is nicer that it is faster now. Could be changed.

Also the speed here is just rough estimation because if player has very high forward/side/up speed, then they will go fast. What are the odds.

Might break existing Noclip% WR.

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -310,6 +310,9 @@ void ServerDLL::Clear()
 	offAngles = 0;
 	offCmd = 0;
 	offBhopcap = 0;
+	offMaxspeed = 0;
+	offClientMaxspeed = 0;
+	offMoveType = 0;
 	pBhopcapWindows = 0;
 	pCZDS_Velocity_Byte = 0;
 	pCBasePlayer__Jump_OldButtons_Check_Byte = 0;
@@ -408,6 +411,9 @@ void ServerDLL::FindStuff()
 			offInDuck = 0x90;
 			offFlags = 0xB8;
 			offBasevelocity = 0x74;
+			offMaxspeed = 0x1f4;
+			offClientMaxspeed = 0x1f8;
+			offMoveType = 0xdc;
 		});
 
 	auto fPM_Jump = FindFunctionAsync(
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2804,16 +2810,41 @@ HOOK_DEF_2(ServerDLL, void, __cdecl, PM_Move, struct playermove_s*, ppmove, int,
...
 	auto flags = reinterpret_cast<int *>(pmove + offFlags);
+	usercmd_t *cmd = reinterpret_cast<usercmd_t*>(pmove + offCmd);
+	float *maxspeed = reinterpret_cast<float*>(pmove + offMaxspeed);
+	float *clientmaxspeed = reinterpret_cast<float*>(pmove + offClientMaxspeed);
 
 	auto start_origin = Vector(origin);
 
+	auto &hw = HwDLL::GetInstance();
+
+	if (hw.noclip_anglehack && *hw.noclip_anglehack) {
```
SmileyAG, line 2821:
```cpp
offMovetype = 0xDC;
```

```cpp
int *movetype = reinterpret_cast<int*>(pmove + offMovetype);
if (*movetype == MOVETYPE_NOCLIP) {
```

I'm going to remove declarations of `noclip_anglehack` once I touch the BXT code again, because it's actually not needed to be honest, it just takes extra time to find it for each build, **MOVETYPE_NOCLIP** check would be better even

SmileyAG (changes requested):

YaLTeR:
> By default noclip is now faster.

Tbh, considering noclip% is a thing (even if a meme), I think by default it should not change the speed. Especially since right now it's not clear how to restore the default behavior. Could you add like an _enabled cvar too?

khanghugo:
ye

khanghugo:
This one has the suggested changes but I squashed all the commits.

YaLTeR:
Weird hlstrafe commit needs to be removed, otherwise good to go

YaLTeR (approved):

YaLTeR:
And I guess resolve conflict

khanghugo:
There is no conflict here.

YaLTeR:
Thanks

PR #508, opened 2024-02-21
Warn if Metamod is detected. Auto disable it for cstrike.

### LINKED ISSUE #501: Give an error and close the game if metamod is enabled
I think there are lots of potentials for BXT to be casually used by mappers because of very helpful HUD info.

The problem now is that most people are relying on AmdModX for basic features like checkpoint or hook or fast noclip to move around.

AmdModX uses MetaMod in between the game and server DLL. So, BXT injection right now would crash the game.

If BunnymodXT is injected, maybe there is a way to force the game to load the normal DLLs. It will be very convenient for most people.

khanghugo:
Closes #501 

"Warn" in here means exiting the game right away. I am not sure how to do it more elegantly.


khanghugo:
Hmmm, it doesn't work on Windows. I keep getting error.
image

String manipulation in C++ is fairly arcane for me.

SmileyAG:
> Hmmm, it doesn't work on Windows. I keep getting error.
> 
> String manipulation in C++ is fairly arcane for me.

Well, the crucial issue with this code is actually not how you did the string manipulation, but how you declared the functions

You can only use `__fastcall` if you declare a member functions for **Windows** (e.g. *CBasePlayer::TakeDamage*)
In all other majority of cases, `__cdecl` will be your choice

Also, when you use `__fastcall`, you will need to specify **int edx** as the second parameter
**edx** doesn't seem to be used, so you can return any value to it (e.g. `0`) in case you need to call that function

Examples:

```cpp
HOOK_DECL(void, __cdecl, PM_PreventMegaBunnyJumping) // Windows, Linux

typedef void(__fastcall *_CBasePlayer__GiveNamedItem)(void *thisptr, int edx, const char *pszName); // Windows
_CBasePlayer__GiveNamedItem ORIG_CBasePlayer__GiveNamedItem;
typedef void(__cdecl *_CBasePlayer__GiveNamedItem_Linux)(void *thisptr, const char *pszName); // Linux
_CBasePlayer__GiveNamedItem_Linux ORIG_CBasePlayer__GiveNamedItem_Linux;
```

```cpp
#ifdef _WIN32
ORIG_CBasePlayer__GiveNamedItem(classPtr, 0, HwDLL::GetInstance().GetString(iszItem));
#else
ORIG_CBasePlayer__GiveNamedItem_Linux(classPtr, HwDLL::GetInstance().GetString(iszItem));
#endif
```

SmileyAG:
I also made a pull request with the necessary changes to that branch: https://github.com/khanghugo/BunnymodXT/pull/4

khanghugo:
Thanks. Looking good.

SmileyAG (approved):

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,51 @@
+#include "stdafx.hpp"
+#include "modules.hpp"
+
+namespace helper_functions
+{
+	void com_fixslashes(std::string &str)
+	{
+		// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/game_shared/bot/nav_file.cpp#L680
+		#ifdef _WIN32
+		str = std::regex_replace(str, std::regex("/"), "\\");
```
YaLTeR, line 10: Does this really need a regex for a single character? Can this maybe use `std::replace`?

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -8053,3 +8071,34 @@ HOOK_DEF_0(HwDLL, qboolean, __cdecl, CL_ReadDemoMessage_OLD)
 
 	return rv;
 }
+
+HOOK_DEF_1(HwDLL, void, __cdecl, LoadThisDll, const char*, szDllFilename)
+{
+	if (boost::ends_with(szDllFilename, helper_functions::add_os_library_extension("metamod").c_str()))
```
YaLTeR, line 8077: I can't look at these two (three?) allocations (one for converting "metamod" to std::string, second for allocating the return std::string from add_os_library_extension, third for converting ".dll" to std::string) with a straight face. Maybe do something like

```
#ifdef _WIN32
#define DLL_EXTENSION ".dll"
#else
#define DLL_EXTENSION ".so"
#endif
```

in that header, then `boost::ends_with(szDllFilename, "metamod" DLL_EXTENSION)`? Since you can concatenate constexpr strings like that.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -8053,3 +8071,34 @@ HOOK_DEF_0(HwDLL, qboolean, __cdecl, CL_ReadDemoMessage_OLD)
...
+	{
+		EngineDevMsg("[hw dll] AmxModX detected.\n");
```
YaLTeR, line 8079: Is it more intuitive to say AmxModX here instead of Metamod?

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	const char *swap_lib(const char* current_lib_path, std::string new_lib_path, const char *start)
+	{
+		com_fixslashes(new_lib_path);
+
+		const std::string filename = current_lib_path;
+		const auto index = filename.find(start);
+		static std::string new_path;
+		new_path = filename.substr(0, index) + add_os_library_extension(new_lib_path).c_str();
```
YaLTeR, line 34: Here's a fun C++ question, does the std::string returned from `add_os_library_extension(new_lib_path)` live long enough for this to be safe? Or is it deallocated right after the call to `c_str()` but before the call to `+` so it's actually broken?

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,51 @@
...
+	{
+		EngineWarning("%s", str.c_str());
+
+		#ifdef _WIN32
+		MessageBox(NULL, str.c_str(), "Fatal Error", MB_OK | MB_ICONERROR);
+		#endif
+
+		ClientDLL::GetInstance().pEngfuncs = nullptr;
+		ServerDLL::GetInstance().pEngfuncs = nullptr;
+		HwDLL::GetInstance().ppGlobals = nullptr;
```
YaLTeR, line 49: Certainly an original way to cause a crash. Maybe `std::exit`?

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+		return str += extension;
+	}
+
+	const char *swap_lib(const char* current_lib_path, std::string new_lib_path, const char *start)
+	{
+		com_fixslashes(new_lib_path);
+
+		const std::string filename = current_lib_path;
+		const auto index = filename.find(start);
```
YaLTeR, line 32: What if it's not found? I.e. ppl put metamod somewhere other than `addons`.

YaLTeR (changes requested):

khanghugo:
Eh, I wouldn't know really. @SmileyAG can explain that.

SmileyAG:
Made appropriate changes from reviews.

And I handled most of edge cases for that request, but let know if I missed something more, he-he.

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,44 @@
...
+		#ifdef _WIN32
+		std::replace(str.begin(), str.end(), '/', '\\');
+		#else
+		std::replace(str.begin(), str.end(), '\\', '/');
+		#endif
+	}
+
+	const char *swap_lib(const char* current_lib_path, std::string new_lib_path, const char *start)
+	{
+		const std::string filename = current_lib_path;
```
YaLTeR, line 19: This already allocates unconditionally; might as well just return the `std::string`? That'll also get rid of the awkward static string.

BunnymodXT/helper_functions.cpp
```diff
@@ -0,0 +1,44 @@
...
+		const auto index = filename.find(start);
+
+		if ((index == std::string::npos) || // String not found in current path.
+			((index > 0) && (filename[index - 1] != PATH_SLASH))) // Previous character from the specified start is not a slash.
+			return current_lib_path;
+
+		com_fixslashes(new_lib_path);
```
YaLTeR, line 26: I'm not sure that's needed; you already pass the right slashed version at the call site.

BunnymodXT/helper_functions.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+namespace helper_functions
+{
+	#ifdef _WIN32
+	#define DLL_EXTENSION ".dll"
+	#define PATH_SLASH '\\'
+	#else
+	#define DLL_EXTENSION ".so"
+	#define PATH_SLASH '/'
+	#endif
```
YaLTeR, line 11: Defines cannot be scoped, so this looks a little misleading. Please move above the namespace.

BunnymodXT/stdafx.hpp
```diff
@@ -46,6 +46,7 @@ typedef unsigned long long steamid_t;
 #include <vector>
 #include <unordered_map>
 #include <utility>
+#include <regex>
```
YaLTeR, line 49: No longer needed
```suggestion
```

BunnymodXT/helper_functions.cpp
SmileyAG: Well, you know, in this particular case the names of the libraries were different (win: `mp.dll`, linux: `cs.so`), so in this case you are right

But what if next time I want to type the path to a library that has the same name on both platforms, and I don't want to duplicate code with using preprocessor flags only to manually change the type of slashes?

So let this function do it for us instead

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -8092,7 +8092,7 @@ HOOK_DEF_1(HwDLL, void, __cdecl, LoadThisDll, const char*, szDllFilename)
 			#endif
 
 			EngineDevMsg("[hw dll] Old path to game library: %s\n", szDllFilename);
-			szDllFilename = helper_functions::swap_lib(szDllFilename, cs_lib, "addons");
+			szDllFilename = helper_functions::swap_lib(szDllFilename, cs_lib, "addons").c_str();
```
YaLTeR, line 8095: This deallocates the returned string right away; the c_str() becomes invalid. You need to store it somewhere on the stack where it will be alive while it is used.

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -8092,7 +8092,8 @@ HOOK_DEF_1(HwDLL, void, __cdecl, LoadThisDll, const char*, szDllFilename)
 			#endif
 
 			EngineDevMsg("[hw dll] Old path to game library: %s\n", szDllFilename);
-			szDllFilename = helper_functions::swap_lib(szDllFilename, cs_lib, "addons").c_str();
+			const char *swap_lib_output = helper_functions::swap_lib(szDllFilename, cs_lib, "addons").c_str();
```
YaLTeR, line 8095: This suffers from the same problem

YaLTeR (approved):
Just to be sure @khanghugo could you test this still works on Linux?

khanghugo:
Works for me.

YaLTeR:
Thanks

PR #510, opened 2024-02-26
feature: landmark for trigger_teleport

khanghugo:
Something from CSS. Something like this https://github.com/s1lentq/ReGameDLL_CS/issues/950

~~I have to hijack the "message" field because `trigger_teleportation` does not have its own key value function.~~

~~I don't really hope this thing is merged but I want it stay around in case anyone wanting to test~~ landmark teleportation in GoldSrc. It is a neat feature, very simple and effective.

Here is the map to test, along with its source 
[arte_confused.zip](https://github.com/YaLTeR/BunnymodXT/files/14409237/arte_confused.zip)

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -7672,6 +7672,7 @@ HOOK_DEF_3(HwDLL, int, __cdecl, SV_SpawnServer, int, bIsDemo, char*, server, cha
 	if (ret) {
 		Interprocess::WriteMapChange(CustomHud::GetTime(), server);
 		lastLoadedMap = server;
+		ServerDLL::GetInstance().ClearTPLandmarks();
 	}
 
 	if (insideHost_Loadgame_f) {
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -7669,10 +7669,12 @@ HOOK_DEF_0(HwDLL, void, __cdecl, R_DrawSkyBox)
 HOOK_DEF_3(HwDLL, int, __cdecl, SV_SpawnServer, int, bIsDemo, char*, server, char*, startspot)
 {
 	auto ret = ORIG_SV_SpawnServer(bIsDemo, server, startspot);
+
+	ServerDLL::GetInstance().ClearTPLandmarks();
+
 	if (ret) {
 		Interprocess::WriteMapChange(CustomHud::GetTime(), server);
 		lastLoadedMap = server;
-		ServerDLL::GetInstance().ClearTPLandmarks();
 	}
 
 	if (insideHost_Loadgame_f) {
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -3504,7 +3504,7 @@ void TriggerTpLandmarkAfter(entvars_t *pev, Vector offset)
 {
 	pev->origin = pev->origin + offset;
 	// have to offset by some HULL because of origin z diff
-	const auto is_duck = pev->button & (IN_DUCK) || pev->flags & (FL_DUCKING);
+	const auto is_duck = pev->bInDuck || pev->flags & (FL_DUCKING);
 	const Vector VEC_HULL_MIN(-16, -16, -36);
 	const Vector VEC_DUCK_HULL_MIN(-16, -16, -18);
 	const auto hull_offset = is_duck ? VEC_DUCK_HULL_MIN : VEC_HULL_MIN;
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -3505,8 +3505,6 @@ void TriggerTpLandmarkAfter(entvars_t *pev, Vector offset)
 	pev->origin = pev->origin + offset;
 	// have to offset by some HULL because of origin z diff
 	const auto is_duck = pev->bInDuck || pev->flags & (FL_DUCKING);
-	const Vector VEC_HULL_MIN(-16, -16, -36);
-	const Vector VEC_DUCK_HULL_MIN(-16, -16, -18);
 	const auto hull_offset = is_duck ? VEC_DUCK_HULL_MIN : VEC_HULL_MIN;
 
 	pev->origin[2] += hull_offset[2];
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -114,6 +114,11 @@ extern "C" void __cdecl _ZN12CBaseTrigger13TeleportTouchEP11CBaseEntity(void* th
 {
 	return ServerDLL::HOOKED_CBaseTrigger__TeleportTouch_Linux(thisptr, pOther);
 }
+
+extern "C" void __cdecl _Z16DispatchKeyValueP7edict_sP14KeyValueData_s(edict_t* pentKeyvalue, KeyValueData* pkvd)
+{
+	return ServerDLL::HOOKED_DispatchKeyValue(pentKeyvalue, pkvd);
+}
 #endif
 
 void ServerDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* moduleBase, size_t moduleLength, bool needToIntercept)
@@ -180,7 +185,8 @@ void ServerDLL::Hook(const std::wstring& moduleName, void* moduleHandle, void* m
 			ORIG_CBaseEntity__IsInWorld, HOOKED_CBaseEntity__IsInWorld,
 			ORIG_CBaseEntity__IsInWorld_Linux, HOOKED_CBaseEntity__IsInWorld_Linux,
 			ORIG_CBaseTrigger__TeleportTouch, HOOKED_CBaseTrigger__TeleportTouch,
-			ORIG_CBaseTrigger__TeleportTouch_Linux, HOOKED_CBaseTrigger__TeleportTouch_Linux);
+			ORIG_CBaseTrigger__TeleportTouch_Linux, HOOKED_CBaseTrigger__TeleportTouch_Linux,
+			ORIG_DispatchKeyValue, HOOKED_DispatchKeyValue);
 	}
 }
 
@@ -237,7 +243,8 @@ void ServerDLL::Unhook()
 			ORIG_CBaseEntity__IsInWorld,
 			ORIG_CBaseEntity__IsInWorld_Linux,
 			ORIG_CBaseTrigger__TeleportTouch,
-			ORIG_CBaseTrigger__TeleportTouch_Linux);
+			ORIG_CBaseTrigger__TeleportTouch_Linux,
+			ORIG_DispatchKeyValue);
 	}
 
 	Clear();
@@ -296,6 +303,7 @@ void ServerDLL::Clear()
 	ORIG_CBaseEntity__FireBullets = nullptr;
 	ORIG_DispatchSpawn = nullptr;
 	ORIG_DispatchTouch = nullptr;
+	ORIG_DispatchKeyValue = nullptr;
 	ORIG_CBaseEntity__FireBullets_Linux = nullptr;
 	ORIG_CBaseEntity__FireBulletsPlayer = nullptr;
 	ORIG_CBaseEntity__FireBulletsPlayer_Linux = nullptr;
@@ -361,6 +369,8 @@ void ServerDLL::Clear()
 	offm_fStamina = 0;
 	offm_old_iAmmo = 0;
 	offm_iPlayerSaveLock = 0;
+
+	tpLandmarks.clear();
 }
 
 bool ServerDLL::CanHook(const std::wstring& moduleFullName)
@@ -1223,6 +1233,7 @@ void ServerDLL::FindStuff()
 
 	ORIG_DispatchSpawn = reinterpret_cast<_DispatchSpawn>(MemUtils::GetSymbolAddress(m_Handle, "_Z13DispatchSpawnP7edict_s"));
 	ORIG_DispatchTouch = reinterpret_cast<_DispatchTouch>(MemUtils::GetSymbolAddress(m_Handle, "_Z13DispatchTouchP7edict_sS0_"));
+	ORIG_DispatchKeyValue = reinterpret_cast<_DispatchKeyValue>(MemUtils::GetSymbolAddress(m_Handle, "_Z16DispatchKeyValueP7edict_sP14KeyValueData_s"));
 	ORIG_ClientCommand = reinterpret_cast<_ClientCommand>(MemUtils::GetSymbolAddress(m_Handle, "_Z13ClientCommandP7edict_s"));
 	ORIG_PlayerPostThink = reinterpret_cast<_PlayerPostThink>(MemUtils::GetSymbolAddress(m_Handle, "_Z15PlayerPostThinkP7edict_s"));
 	ORIG_PM_Move = reinterpret_cast<_PM_Move>(MemUtils::GetSymbolAddress(m_Handle, "PM_Move"));
@@ -1230,10 +1241,11 @@ void ServerDLL::FindStuff()
 	ORIG_CmdStart = reinterpret_cast<_CmdStart>(MemUtils::GetSymbolAddress(m_Handle, "_Z8CmdStartPK7edict_sPK9usercmd_sj"));
 	ORIG_CmdEnd = reinterpret_cast<_CmdEnd>(MemUtils::GetSymbolAddress(m_Handle, "_Z6CmdEndPK7edict_s"));
 
-	if (ORIG_DispatchSpawn && ORIG_DispatchTouch && ORIG_ClientCommand && ORIG_PlayerPostThink && 
+	if (ORIG_DispatchSpawn && ORIG_DispatchTouch && ORIG_DispatchKeyValue && ORIG_ClientCommand && ORIG_PlayerPostThink && 
 		ORIG_PM_Move && ORIG_AddToFullPack && ORIG_CmdStart && ORIG_CmdEnd) {
 		EngineDevMsg("[server dll] Found DispatchSpawn at %p.\n", ORIG_DispatchSpawn);
 		EngineDevMsg("[server dll] Found DispatchTouch at %p.\n", ORIG_DispatchTouch);
+		EngineDevMsg("[server dll] Found DispatchKeyValue at %p.\n", ORIG_DispatchKeyValue);
 		EngineDevMsg("[server dll] Found ClientCommand at %p.\n", ORIG_ClientCommand);
 		EngineDevMsg("[server dll] Found PlayerPostThink at %p.\n", ORIG_PlayerPostThink);
 		EngineDevMsg("[server dll] Found PM_Move at %p.\n", ORIG_PM_Move);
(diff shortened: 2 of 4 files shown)


khanghugo:
The only reason why I say that "I don't really hope this thing is merged" because the implementation is very constrained. I cannot really think of a way to make maps taking advantage of this to just work. That means mappers must be aware of BXT to add value into "message" field or somebody else has to edit the map with bspguy or something. Also simply hijacking the "message" field is not very nice (attitude).

SmileyAG:
> That means mappers must be aware of BXT to add value into "message" field or somebody else has to edit the map with bspguy or something.

I suppose `DispatchKeyValue` function can be hooked from exported table (`funcs.pfnKeyValue`) for hijacking keyvalues:
https://github.com/YaLTeR/BunnymodXT/blob/a0cc46992a10612e1253d4ac2b146babbf383438/HLSDK/engine/eiface.h#L424

Then perform all needed conditions, if result is positive then received edict and its keyvalue we can store all in our custom container at the same index (e.g. `std::vector` with `emplace_back`)

https://github.com/YaLTeR/BunnymodXT/blob/a0cc46992a10612e1253d4ac2b146babbf383438/HLSDK/engine/eiface.h#L297-L304

In our function we can iterate over that container and check whether the `szKeyName` matches with what you expect and whether the edict from the same index in the container matches the one edict you are checking for
If all matches, then extract data from `szValue` and apply it in your code

In the end, clean the container every time the server is shutdown (`ServerDeactivate` function should be ideal for this task, it is also exported by using `funcs.pfnServerDeactivate`):
https://github.com/YaLTeR/BunnymodXT/blob/a0cc46992a10612e1253d4ac2b146babbf383438/HLSDK/engine/eiface.h#L446-L448

khanghugo:
Thanks for the suggestion, I think this should be better.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		pev->v_angle = prev_view;
+		pev->angles = prev_angles;
+	}
+}
+
+#define STRING(offset)		((const char *)(HwDLL::GetInstance().ppGlobals->pStringBase + (unsigned int)(offset)))
+
+HOOK_DEF_2(ServerDLL, void, __cdecl, DispatchKeyValue, edict_t*, pentKeyvalue, KeyValueData*, pkvd)
+{
+	if (pkvd && pentKeyvalue
```
SmileyAG, line 3428:
```cpp
	if (pkvd && pentKeyvalue && pkvd->szClassName && pkvd->szKeyName && pkvd->szValue)
	{
		if (!strcmp(pkvd->szClassName, "trigger_teleport") && !strcmp(pkvd->szKeyName, "landmark"))
		{
			const auto index = pEngfuncs->pfnIndexOfEdict(pentKeyvalue);
			tpLandmarks[index] = pkvd->szValue;
		}
	}
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		&& !strcmp(pkvd->szClassName, "trigger_teleport") 
+		&& !strcmp(pkvd->szKeyName, "landmark")
+		) {
+
+		edict_t *edicts;
+		HwDLL::GetInstance().GetEdicts(&edicts);
+		const auto index = pentKeyvalue - edicts;
+
+		tpLandmarks[index] = pkvd->szValue;
+		pkvd->fHandled = 1;
```
SmileyAG, line 3439: Don't set `pkvd->fHandled = 1`!

Because in some game it may turn out that in the KeyValue function of the corresponding class this case will be handled, then you simply will not allow the game to read data from it, I think the consequences are not worth explaining

I think we should add the following comment at the beginning of the function:

```cpp
// Don't set pkvd->fHandled = 1, as some games could handle the same cases for the same entities, which means you won't give the game a chance to read that data!
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		EngineDevWarning("handled!!!!\n");
+	}
+
+	ORIG_DispatchKeyValue(pentKeyvalue, pkvd);
+}
+
+// std::optional is at least C++17 :DDDDDDDDDDDDDDDDDDDDDDD
+std::tuple<bool, Vector> TriggerTpLandmarkBefore(bool work, entvars_t *this_pev, entvars_t *pev, enginefuncs_t *pEngfuncs)
```
SmileyAG, line 3447:
```cpp
std::tuple<bool, Vector> TriggerTpLandmarkBefore(bool enabled, entvars_t *this_pev, entvars_t *pev)
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+
+	return std::make_tuple(false, Vector());
+}
+
+void TriggerTpLandmarkAfter(entvars_t *pev, Vector offset)
+{
+	pev->origin = pev->origin + offset;
+	// have to offset by some HULL because of origin z diff
+	auto is_duck = pev->button & (IN_DUCK) || pev->flags & (FL_DUCKING);
+	if (is_duck)
```
SmileyAG, line 3480:
```cpp
const Vector VEC_HULL_MIN(-16, -16, -36);
const Vector VEC_DUCK_HULL_MIN(-16, -16, -18);
auto hull_offset = is_duck ? VEC_DUCK_HULL_MIN : VEC_HULL_MIN;

pev->origin[2] += hull_offset[2];
```

https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/util.h#L414-L424

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+	HwDLL::GetInstance().GetEdicts(&edicts);
+	int index = this_pev->pContainingEntity - edicts;
+	const auto tpLandmarks = ServerDLL::GetInstance().tpLandmarks;
+
+	if (!tpLandmarks.count(index))
+		return std::make_tuple(false, Vector());
+
+	auto what = tpLandmarks.at(index);
+	edict_t *landmark = pEngfuncs->pfnFindEntityByString(NULL, "targetname", what.c_str());
+	auto available = true;
```
SmileyAG, line 3462:
```cpp
if (HwDLL::GetInstance().IsValidEdict(landmark))
	return std::make_tuple(true, Vector(pev->origin) - Vector(landmark->v.origin));
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+
+	edict_t *edicts;
+	HwDLL::GetInstance().GetEdicts(&edicts);
+	int index = this_pev->pContainingEntity - edicts;
+	const auto tpLandmarks = ServerDLL::GetInstance().tpLandmarks;
+
+	if (!tpLandmarks.count(index))
+		return std::make_tuple(false, Vector());
+
+	auto what = tpLandmarks.at(index);
```
SmileyAG, line 3460:
```cpp
auto landmarkName = tpLandmarks.at(index);
edict_t *landmark = ServerDLL::GetInstance().pEngfuncs->pfnFindEntityByString(NULL, "targetname", landmarkName.c_str());
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		int index = pEngfuncs->pfnIndexOfEdict(ent);
+
+		if ((index < 1) || (index > hw.ppGlobals->maxClients)) // gGlobalVariables.maxClients = svs.maxclients
+			return false;
+	}
+
+	return true;
+}
+
+void TriggerTpKeepsMomentumRestore(Vector prev_vel, Vector prev_view, Vector prev_angles, Vector prev_basevelocity, entvars_t *pev, enginefuncs_t *pEngfuncs)
```
SmileyAG, line 3385:
```cpp
void TriggerTpKeepsMomentumRestore(entvars_t *pev, Vector prev_vel, Vector prev_view, Vector prev_angles, Vector prev_basevelocity)
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+	ORIG_DispatchKeyValue(pentKeyvalue, pkvd);
+}
+
+// std::optional is at least C++17 :DDDDDDDDDDDDDDDDDDDDDDD
+std::tuple<bool, Vector> TriggerTpLandmarkBefore(bool work, entvars_t *this_pev, entvars_t *pev, enginefuncs_t *pEngfuncs)
+{
+	if (!work)
+		return std::make_tuple(false, Vector());
+
+	edict_t *edicts;
```
SmileyAG, line 3452:
```cpp
const auto index = ServerDLL::GetInstance().pEngfuncs->pfnIndexOfEdict(this_pev->pContainingEntity);
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+{
+	// Set velocity before viewangles because viewangles will mess with the velocity angle for redirection
+	if (CVars::bxt_ch_trigger_tp_keeps_momentum_velocity.GetBool()) {
+		if (CVars::bxt_ch_trigger_tp_keeps_momentum_velocity_redirect.GetBool()) {
+			// https://github.com/fireblizzard/agmod/blob/bf06e4ffd31c1427784685118820e15552803bcb/dlls/triggers.cpp#L1935
+			// After teleportation, pevToucher has the same viewangles as pentTarget.
+			Vector vecAngles = Vector(0, pev->v_angle.y, 0);
+			Vector vecForward;
+
+			pEngfuncs->pfnAngleVectors(vecAngles, vecForward, nullptr, nullptr);
```
SmileyAG, line 3395:
```cpp
ServerDLL::GetInstance().pEngfuncs->pfnAngleVectors(vecAngles, vecForward, nullptr, nullptr);
```

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+
+	const auto landmark_info = TriggerTpLandmarkBefore(is_bxt_ch_trigger_tp_landmark_enabled, this_pev, pev, pEngfuncs);
+
+	ORIG_CBaseTrigger__TeleportTouch(thisptr, edx, pOther);
+
+	if (is_bxt_ch_trigger_tp_keeps_momentum_enabled && pev && pEngfuncs && IsPlayer(pev->pContainingEntity)) {
+		TriggerTpKeepsMomentumRestore(prev_vel, prev_vel, prev_angles, prev_basevelocity, pev, pEngfuncs);
+	}
+
+	EngineDevWarning("enabled %d pEngfuncs is %d\n", is_bxt_ch_trigger_tp_landmark_enabled, pEngfuncs == NULL);
```
SmileyAG, line 3518: Remove those debug warnings.

SmileyAG (changes requested):

SmileyAG:
> Apparently this does not work on Windows atm. Not sure why.

The reason why it does not work on Windows is because the hook was not declared in the `MemUtils::Intercept` and `MemUtils::RemoveInterception` blocks

https://github.com/YaLTeR/BunnymodXT/blob/76aefb29543951ab4ef3eb90084d7add4f098f88/BunnymodXT/modules/ServerDLL.cpp#L128
https://github.com/YaLTeR/BunnymodXT/blob/76aefb29543951ab4ef3eb90084d7add4f098f88/BunnymodXT/modules/ServerDLL.cpp#L183

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -7529,6 +7529,7 @@ HOOK_DEF_3(HwDLL, int, __cdecl, SV_SpawnServer, int, bIsDemo, char*, server, cha
 	if (ret) {
 		Interprocess::WriteMapChange(CustomHud::GetTime(), server);
 		lastLoadedMap = server;
+		ServerDLL::GetInstance().ClearTPLandmarks();
```
SmileyAG, line 7532: Ugh. You know, clearing in `SV_SpawnServer` is generally not bad, because it is called when a new map starts, changelevels and similar events, and it is called before the engine calls `DispatchKeyValue`

`ServerDeactivate` (**funcs.pfnServerDeactivate** from `DLL_FUNCTIONS` table) is also called in those events and also before `DispatchKeyValue`, but the difference is that it also covers other events such as disconnect
At that point clearing in `ServerDeactivate` looks a bit safer, because for example if you are disconnected and you are in the menu, you know that the data will already be cleared, and in the case of `SV_SpawnServer` you need load map to clear this old data

So personally, I would advise hooking `ServerDeactivate` from the server-side (fortunately it doesn’t require patterns) and including a comment at the beginning of its function about the difference in data cleanup compared to `SV_SpawnServer`

You may also need to hook `ServerActivate` to track its state, because HLSDK does this to avoid unnecessary calls to `ServerDeactivate`, take a look at the highlighted lines: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/client.cpp#L685-L708
`ServerActivate` could be hooked from exported server table too, so not need to worries about patterns

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3316,4 +3359,199 @@ HOOK_DEF_1(ServerDLL, int, __cdecl, CBaseEntity__IsInWorld_Linux, void*, thisptr
...
+		return std::make_tuple(true, Vector(pev->origin) - Vector(landmark->v.origin));
+
+	return std::make_tuple(false, Vector());
+}
+
+void TriggerTpLandmarkAfter(entvars_t *pev, Vector offset)
+{
+	pev->origin = pev->origin + offset;
+	// have to offset by some HULL because of origin z diff
+	auto is_duck = pev->button & (IN_DUCK) || pev->flags & (FL_DUCKING);
```
SmileyAG, line 3479: Maybe this check should be next?

```cpp
auto is_duck = pev->bInDuck || pev->flags & (FL_DUCKING);
```

Because checking for a pressed key with `+duck` input does not look completely reliable
In some games, duck on the movement code side may be disabled, but the command for duck input is available in the game

`bInDuck` is set during the crouching process, and `FL_DUCKING` flag is set when we are definitely crouched

BunnymodXT/modules/ServerDLL.cpp
khanghugo: So many repetition of this HULL values, maybe future refactor?

BunnymodXT/modules/ServerDLL.cpp
khanghugo: This is rebase of another other commit, maybe ill change it there instead

BunnymodXT/modules/ServerDLL.cpp
khanghugo: Rebase of another commit

BunnymodXT/modules/ServerDLL.cpp
khanghugo: Eh, I am not sure what else to check here for duck beside those two. On top of that, the condition for landmark tp to work is very involving and the mapper (or map editor) has to allow it to happen. It is possible that a different game mod will implement this inside their game mod already without relying on this.

BunnymodXT/modules/HwDLL.cpp
khanghugo: What do you mean by tracking `ServerActivate` state? Tracking `g_serveractive`? So there needs to be either another pattern for that static value or BXT own static value for `g_serveractive`? Seems like a lot of extra work.  I think clearing data on load is fine unless you are adamant about it then I can make some changes for that.

BunnymodXT/modules/ServerDLL.cpp
SmileyAG: > So many repetition of this HULL values, maybe future refactor?

Surely.

I know that now in the code these HULL values are declared several times in scope of different functions, but with a future refactor all this will be declared once in some header.

So let it be like this for now.

BunnymodXT/modules/HwDLL.cpp
SmileyAG: If you read the highlighted lines carefully, you will understand that you don’t need to intercept that `g_serveractive` variable at all. Let me show you a concept of how this can be done without patterns.


```cpp
// Global variable
bool g_serveractive = false;

// Find "DLL_FUNCTIONS funcs;" in the server module for the declaration location
ORIG_ServerActivate = funcs.pfnServerActivate;
ORIG_ServerDeactivate = funcs.pfnServerDeactivate;

HOOK_DEF_0(ServerDLL, void, __cdecl, ServerDeactivate)
{
	// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/client.cpp#L685-L708
	if (g_serveractive)
	{
		g_serveractive = false;

		// Peform any shutdown operations here...
	}

	ORIG_ServerDeactivate();
}

HOOK_DEF_3(ServerDLL, void, __cdecl, ServerActivate, edict_t*, pEdictList, int, edictCount, int, clientMax)
{
	g_serveractive = true;
	ORIG_ServerActivate(pEdictList, edictCount, clientMax);
}
```

SmileyAG: > on load is fine unless you are adamant about it then I can make some changes for that.

Well I'd fine with it in general because I don't think there would a cases when you want to access that data while being not on the server, and even so there seems a validation checks is done, so I'd like to say fine, but `ServerDeactivate` would be a nicer option in fact, maybe it should be done with future refactoring then

khanghugo: I totally understood what you mean. I just wonder if it is worth it. Do you think it is good to pursue this?

khanghugo: Github going slow. I agree it should be in the future refactor. This PR is another rebase of another PR so I am not very confident to make some changes here.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3439,10 +3454,76 @@ void TriggerTpKeepsMomentumRestore(Vector prev_vel, Vector prev_view, Vector pre
 	}
 }
 
+HOOK_DEF_2(ServerDLL, void, __cdecl, DispatchKeyValue, edict_t*, pentKeyvalue, KeyValueData*, pkvd)
+{
+	if (pkvd && pentKeyvalue
+		&& pkvd->szClassName // some entities don't have classname, this avoids crash.
+		&& pkvd->szKeyName && pkvd->szValue // just to make sure
+		&& !strcmp(pkvd->szClassName, "trigger_teleport") 
+		&& !strcmp(pkvd->szKeyName, "landmark")
```
YaLTeR, line 3463: That's a lot of string comparisons that happen all the time (?), especially for an optional feature.

BunnymodXT/modules/ServerDLL.cpp
YaLTeR: I have the same question as Smiley

BunnymodXT/modules/ServerDLL.cpp
khanghugo: You made me paranoid so I debug print it. This is only called when a new server is started. Every time a new server is started, a lot of string comparisons like this happen to parse entities in BSP. This one is just one extra entity available in the game.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -3439,10 +3454,76 @@ void TriggerTpKeepsMomentumRestore(Vector prev_vel, Vector prev_view, Vector pre
...
+	return std::make_tuple(false, Vector());
+}
+
+void TriggerTpLandmarkAfter(entvars_t *pev, Vector offset)
+{
+	pev->origin = pev->origin + offset;
+	// have to offset by some HULL because of origin z diff
+	const auto is_duck = pev->bInDuck || pev->flags & (FL_DUCKING);
+	const Vector VEC_HULL_MIN(-16, -16, -36);
+	const Vector VEC_DUCK_HULL_MIN(-16, -16, -18);
```
YaLTeR, line 3508: I don't have an update PR button for some reason, but this probably needs to be removed after rebase to master since they are now defines.

YaLTeR (approved):
Otherwise good to go ig

YaLTeR:
Thanks

PR #512, opened 2024-03-03
More structures / constants / enums from the engine and HLSDK

### LINKED ISSUE #486: Add a comments indicating at what point the structures no longer match the Quake code
Example:

```cpp
typedef struct cmd_function_s
{
	struct cmd_function_s	*next;
	char					*name;
	xcommand_t				function;
	
	// Structure is exactly the same as in the Quake code and latest GoldSrc engine up to this point.
	// Therefore, everything added below that comment, for example, may not be exist in older versions of the GoldSrc engine!
	
	int						flags;
} cmd_function_t;
```

Since this with a huge chance guarantees you the fact that Valve did not touch anything at the beginning of these structures in all GoldSrc versions, which means if you have any issues with using the structure in older versions, because this will the narrow down the search for issue and guide you to fix it likely in next steps: inherit a structure / use preprocessor flags, add new member variables corresponding to that engine version and use it if the right version for it

SmileyAG:
Closes #486 

Perhaps I missed also some valued structures / constants and enums from the engine and HLSDK that will definitely come in handy, but it can be added as needed later, so for now this should be more than enough.


HLSDK/cl_dll/hud.h
```diff
@@ -20,7 +20,11 @@
 // CHud handles the message, calculation, and drawing the HUD
 //
 
-// YaLTeR - Heavily stripped down to contain only useful stuff.
+#if !defined ( HUDH )
+#define HUDH
+#ifdef _WIN32
+#pragma once
+#endif
```
YaLTeR, line 27: Can't you just use `#pragma once`? What's the point of both header guards and `#pragma once` at the same time?

BunnymodXT/modules/HwDLL.cpp
YaLTeR, line 1: I think it should still be included in the main build, no harm in that. Maybe make it an underscored command. The less defines and different build types the better.

BunnymodXT/runtime_data.cpp
```diff
@@ -17,6 +17,12 @@
 #include <cereal/types/vector.hpp>
 #include <cereal/archives/binary.hpp>
 
+/*
+	ATTENTION!
+	ALWAYS ADD SAVE NEW DATA TYPES AT THE END OF THE DATA!
+	SO YOU WILL NOT BROKE COMPATIBILITY WITH PARSING OLD DEMOS!
```
YaLTeR, line 23:
```suggestion
	Attention! New runtime data types must be added AT THE END, to keep compatibility with parsing old demos.
```

YaLTeR (commented):
This is a lot of great work!

HLSDK/cl_dll/hud.h
SmileyAG: There is actually a good reason for this.

You see, `#pragma once` was added to GCC starting with version 3.4, which was released in 2004: http://gcc.gnu.org/bugzilla/show_bug.cgi?id=11569
At the same time, MSVC supported `#pragma once` at least as early from 1998 (perhaps even earlier!)

If you want to include these headers in the SDK and compile on a ancient version of GCC, then `#pragma once` would become an issue
That’s why you also need header guards for this case, and `#pragma once` would only as option for Windows

YaLTeR: But this is BXT code, we only support recent compilers?

SmileyAG: > But this is BXT code, we only support recent compilers?

That's true, but if I want to copy-paste that headers to HLSDK and then compile on ancient GCC version?
I want to minimize the difference so that it can be used with most compiler versions, adding three lines with `#ifndef`, `#define` and `#endif` doesn't sounds as complex task, but you guarantee flexibility in most cases with this

YaLTeR: > if I want to copy-paste that headers to HLSDK and then compile on ancient GCC version?

Why would you want to do that?

SmileyAG: Because you want to compile the SDK here and right now, without wasting time updating it to be compatible with the new version of the toolchain

In addition, you want to have ASM instructions as close as possible to the original library, so you must stick with the same version of the compilers that used for the original library

Since it is assumed that in newer compilers their developers could do a lot of internal optimizations, the behavior of which may not be changed even by toggling some flags at the compilation stage (simply: hardcoded), which in the ends affects to what patterns will be for the same code compiled with different versions of toolchain

It is clear that this approach is not suitable for an active project, but it is important to emphasize that I am primarily speaking from the point of view of reverse-engineering
That is, you don’t have the task of updating the toolchain at all; instead, you need to compile the source code of this SDK with full debug information
To then import into Ghidra and, for example, have offsets for all data members of some class

Or in the original library there might not have been a `.pdb` file next to the `.dll` or the `.so` library was compiled with a low level of DWARF information, which is result in less complete backtrace

Here is an example that is not so rare to come across:

You want to make your mod (e.g. Glitchless mod) based on some unique modification, support for which is already done in the BXT side, and therefore you do not decide to update to a new version of the toolchain to not make new support in BXT, because this will most likely entail in the changed patterns and as a result there will be no support (even if we use the same optimization flags as we used for the old toolchain, but there may not be flags for toggling new compiler optimizations)
Of course, in the future we should plan to intercept symbols in the code for all the functions we use, even those that are not exported by default, in case the developer, for his part, decides to export them, and through some global variables with the data type of our custom structure then export them so that we can intercept them on our side, which in end would help us provide support for some functionality even easier (e.g. in the form of their size/offset/name and etc)
Which, simply put, don't force to dependence on the specific toolchain version, realizing that if you (mod developer) update it, support will still be remain, because we (both of projects) have implemented it at a more global level

Perhaps I may moved in topic a bit away from the initial question, but returning to it:

If you can maintain overall compatibility with other versions by minimal effort, then I see no reason not to do so (especially if their implementation does not cause more potential issues overall and these solutions considered as practice standard)

Therefore, if I want to copy a headers from BXT (because as we know, Valve does not provide all the information in SDK, so you have to reverse-engineering some structures to deal with it)
And to use it in a mod that in the original for Linux compiles on a ancient version of GCC, then I will not have issues with compilation due to the use of `#pragma once` only in the case of compiling on Windows and as alternative for Linux I also provide header guards, which in fact is more considered practice standard indeed

YaLTeR: Well, alright. Then can you remove `#pragma once`? It doesn't make sense in presence of include guards since they already do the same thing.

YaLTeR:
Thanks!

PR #514, opened 2024-03-10
Pattern-less implementation of bxt_render_far_entities

### LINKED ISSUE #497: Pattern-less implementation of bxt_render_far_entities
Previously I had done byte patching and set of ent variables to achieve that in BXT code (see all references to `bxt_render_far_entities` in **AddToFullPack** code)

Looking back now I understand that it was bad decision, since it could be done more simpler, it would be supported for each modification and the same end result

The simplified code:
```cpp
// At the start of AddToFullPack function
if (CVars::bxt_render_far_entities.GetBool())
	pSet = NULL;
	
// Rest code
	
auto ret = ORIG_AddToFullPack(state, e, ent, host, hostflags, player, pSet);
	
// Rest code

return ret;
```

Additionally see the comments from that highlighted code: https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/client.cpp#L1114-L1122

SmileyAG:
Closes #497

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -295,7 +295,6 @@ void ServerDLL::Clear()
 	offBhopcap = 0;
 	pBhopcapWindows = 0;
 	pCZDS_Velocity_Byte = 0;
-	pAddToFullPack_PVS_Byte = 0;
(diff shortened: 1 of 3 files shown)


YaLTeR:
Thanks

PR #516, opened 2024-03-16
More helper functions and first steps towards refactoring

SmileyAG:


SmileyAG:
This request, in addition to new helper functions, also generally presents itself as one of the parts of the refactoring, in general it is more or less organized by commits, so I'm really not going to split this request into several small new ones, I hope this is more or less clear now

I have finally decided on the concepts that I want to see, so I don't see more room for improving this request, if you still have any suggestions on what can be improved - then let me know

Few functions such as `build_number` are not used anywhere in the code, but I still prefer to have them on hand for tests and possible new concepts with them in the future, in any case, there are only a few such unused functions in the helper module, so let it be here

SmileyAG:
I have not fully tested this branch yet, so there is no need to merge it yet, but it is now open for review, so you can post your reviews now

SmileyAG:
I am closing this request because it was still something like a testing ground and as a result it is becoming difficult to review, so I will simply divide changes from that branch into approximately 10-20 new pull requests with minor changes.

PR #519, opened 2024-03-30
Add bxt-rs camera replay related hooks

khanghugo:
For https://github.com/YaLTeR/bxt-rs/pull/99

Though might need #518 merged first because this uses variables from there.

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -41,6 +41,7 @@ struct on_tas_playback_frame_data {
 	std::array<float, 4> prev_predicted_trace_fractions;
 	std::array<float, 4> prev_predicted_trace_normal_zs;
 	on_tas_playback_frame_max_accel_yaw_offset max_accel_yaw_offset;
+	std::array<float, 3> rendered_viewangles;
 };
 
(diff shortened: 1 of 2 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -32,6 +32,7 @@ struct on_tas_playback_frame_data {
 	unsigned strafe_cycle_frame_count;
 	std::array<float, 4> prev_predicted_trace_fractions;
 	std::array<float, 4> prev_predicted_trace_normal_zs;
+	std::array<float, 3> rendered_viewangles;
 };
 
 // Change the variable name if you change the parameters!
```
YaLTeR, line 38: I guess this is the time when we need to change the name. I already forgot what I meant by "variable", but we can change the function name to `bxt_on_tas_playback_frame_v2` since the ABI changes with the addition of a field in the struct.

khanghugo: Right, as soon as https://github.com/YaLTeR/BunnymodXT/pull/521 is merged then I can make changes to this (without weird conflicts)

YaLTeR:
Thanks

PR #521, opened 2024-04-02
Add bxt-rs max acceleration yaw offset related hooks

khanghugo:


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -32,6 +32,7 @@ struct on_tas_playback_frame_data {
 	unsigned strafe_cycle_frame_count;
 	std::array<float, 4> prev_predicted_trace_fractions;
 	std::array<float, 4> prev_predicted_trace_normal_zs;
+	std::array<float, 5> accelerated_yawspeed;
```
YaLTeR, line 35: Maybe put as individual fields? Or, even better, make a struct with them and put the struct.

khanghugo: I will be honest. Tuple brainrots me.

khanghugo: This might work for C++ but in Rust I have troubles using struct to handle mut ref. I believe it is skill issue from my side.

YaLTeR: I'll comment on the mut ref, but this value is never used as a ref is it? Just by value

khanghugo: Just values yea. The real horror would be in bxt-rs implementation. But now I think about it named field for enum would hold mut ref just fine, no need for a dedicated struct. I did not realize that that until I use Rust a bit more since then.

YaLTeR:
Guess this needs a hlstrafe update then good to go.

khanghugo:
There will be merge conflict when https://github.com/YaLTeR/BunnymodXT/pull/518 is merged but now is not the time for fear, that comes later.

YaLTeR:
Seems like hlstrafe is still pointing at some other commit

YaLTeR:
Thanks

PR #523, opened 2024-04-06
wrappers: reverse-engineered TriAPI functions from hardware engine

SmileyAG:
This will not give errors after #516 and #512 are merged and after that the rebase of this branch is done

Later I’ll make a test branch based on this with already hooked functions to test the wrappers in action and leave comment about it here, so don't need to merge it yet, but it's open for review

About the function that used from Mesa repository: the main modules (where the `InvertMatrix` was taken) are signed under the **MIT License**, so there should be no difficulties with this: https://docs.mesa3d.org/license.html

In GoldSrc engine, **InvertMatrix** called at the end of `R_SetupGL` function to retrieve **gScreenToWorld** from **gWorldToScreen**:

```cpp
// In the end of R_SetupGL function from engine
InvertMatrix(gWorldToScreen, gScreenToWorld);
```


SmileyAG:
I will squash all the commits from this request to one and rebase it to the master after https://github.com/YaLTeR/BunnymodXT/pull/530 is merged, for now, let it be untouched for a little longer.

PR #526, opened 2024-04-30
bxt_ch_checkpoint: fix stamina for CS 1.6 and restore gravity

### LINKED ISSUE #525: `is_cstrike_dir` is never set for checkpoint go to
https://github.com/YaLTeR/BunnymodXT/blob/76aefb29543951ab4ef3eb90084d7add4f098f88/BunnymodXT/modules/HwDLL.cpp#L3416-L3418

The variable is only set when loading a TAS file, not upon starting the game or starting a map.

https://github.com/YaLTeR/BunnymodXT/blob/76aefb29543951ab4ef3eb90084d7add4f098f88/BunnymodXT/modules/HwDLL.cpp#L6795

Maybe we should have a way to know the current game dir as soon as the game starts so stuffs like this don't scatter around.

khanghugo:
Closes #525 

This might just be a bandaid solution because these values are used many wheres.

Also I happen to play a map with gravity and it doesn't restore my gravity before so here it is.

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -834,6 +834,7 @@ void HwDLL::Clear()
 	ch_checkpoint_vel.clear();
 	ch_checkpoint_viewangles.clear();
 	ch_checkpoint_is_duck.clear();
+	ch_checkpoint_gravity.clear();
 
 
 	tas_editor_mode = TASEditorMode::DISABLED;
@@ -3314,6 +3315,7 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 	static void handler()
 	{
 		auto &hw = HwDLL::GetInstance();
+		auto &cl = ClientDLL::GetInstance();
 
 		auto pl = hw.GetPlayerEdict();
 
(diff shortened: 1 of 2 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -3340,7 +3342,10 @@ struct HwDLL::Cmd_BXT_CH_CheckPoint_Create
 		hw.ch_checkpoint_vel.emplace_back(pl->v.velocity);
 		hw.ch_checkpoint_viewangles.emplace_back(pl->v.v_angle);
 		hw.ch_checkpoint_is_duck.emplace_back(is_duck);
+		hw.ch_checkpoint_gravity.emplace_back(pl->v.gravity);
 		hw.ch_checkpoint_is_set = true;
+		
+		hw.is_cstrike_dir = cl.DoesGameDirMatch("cstrike") || cl.DoesGameDirMatch("czero");
```
YaLTeR, line 3348: Is there really no way to avoid this every time `bxt_ch_checkpoint_create` is called?

khanghugo: Very simple. Hook some functions called when the game is started or something. On similar note, as of now, BXT still does not hook CL_Disconnect, LOL

YaLTeR: Wasn't there some other place in BXT which already had a check like is_cstrike?

khanghugo: Yes but it is only called when a TAS file is loaded, LOL

YaLTeR:
Thanks

PR #528, opened 2024-05-07
minor: added GET_PEV macro for convenient and quick declaration of pev

SmileyAG:


BunnymodXT/shared.hpp
```diff
@@ -45,3 +47,6 @@ constexpr int kRenderFxTrigger = 241; // DO NOT CHANGE THIS VALUE OR YOU WILL BR
 
 // - Other constants
 constexpr steamid_t STEAMID64_CONST = 76561197960265728; // 0x110000100000000
+
+// - Custom macros
+#define GET_PEV(pev, thisptr) entvars_t *pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(thisptr) + off_pev);
```
YaLTeR, line 52: It's better to leave the declaration to the caller.
```suggestion
#define GET_PEV(thisptr) *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(thisptr) + off_pev);
```
Then use like `auto pev = GET_PEV(something);`. This will also fix weirdness like `const GET_PEV(pevToucher, pOther);`

PR #530, opened 2024-07-14
wrappers: a series of engine wrappers has been initialized for support functionality in all builds

SmileyAG:
The wrappers were also divided into separate components for easier navigation and management, since I don’t like it when everything is in one basket.

Content for them will appear with the following pull requests, this request is intended to resolve the issue of what they should look like.

### DIFF
--- a/BunnymodXT/wrappers/engine.cpp
+++ b/BunnymodXT/wrappers/engine.cpp
@@ -1,0 +1,5 @@
+#include "engine.hpp"
+
+namespace Engine
+{
+};
--- a/BunnymodXT/discord_integration.hpp
+++ b/BunnymodXT/discord_integration.hpp
@@ -8,6 +8,7 @@
 #include "modules/HwDLL.hpp"
 #include "cvars.hpp"
 #include "hud_custom.hpp"
+#include "wrappers.hpp"
 
 using namespace std::literals;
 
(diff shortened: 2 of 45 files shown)


BunnymodXT/wrappers/engine.cpp
```diff
@@ -0,0 +1,5 @@
+#include "engine.hpp"
+
+namespace Engine
+{
+};
```
YaLTeR, line 5: Namespaces in .cpp have trailing ; but in .hpp they don't?

YaLTeR (commented):
At a glance looks good; I'd like to see some example of what you put here before merging though. Other than the TriApi one from the other PR.

BunnymodXT/wrappers/engine.cpp
SmileyAG: In general, semicolon at the end for the named namespace is not required by standards. But if we look at the code design from other files of this project, it is clear that someone will put them both in the header and in .cpp, someone decides to put them only in the header, somewhere they are simply not there. So it's all a matter of taste.

But if we still want to adhere to some consistency in the code design, then I think it's easier to remove them altogether. What do you think about this?

YaLTeR: Yeah, let's do that if they are not needed

PR #533, opened 2024-07-14
restructuring: added macros for stringification

SmileyAG:
Most common `CONCAT`, `TOSTRING` and `__func__` macros, but wrapped in their own cover for increased compatibility and avoidance of unnecessary collisions.

Necessary for future pull requests to making bizzare macros in modules to simplify coding in them.

### DIFF
--- a/BunnymodXT/shared.hpp
+++ b/BunnymodXT/shared.hpp
@@ -45,3 +45,11 @@ constexpr int kRenderFxTrigger = 241; // DO NOT CHANGE THIS VALUE OR YOU WILL BR
 
 // - Other constants
 constexpr steamid_t STEAMID64_CONST = 76561197960265728; // 0x110000100000000
+
+// - Custom macros
+
+#define _BXT_STRINGIFY(x) #x
+#define BXT_TOSTRING(x) _BXT_STRINGIFY(x)
+#define BXT_CONCATENATE(x, y) x##y
+#define BXT_CONCATENATE_DETAIL(x, y) BXT_CONCATENATE(x, y)
+#define BXT_FUNC __func__


BunnymodXT/shared.hpp
```diff
@@ -45,3 +45,11 @@ constexpr int kRenderFxTrigger = 241; // DO NOT CHANGE THIS VALUE OR YOU WILL BR
 
 // - Other constants
 constexpr steamid_t STEAMID64_CONST = 76561197960265728; // 0x110000100000000
+
+// - Custom macros
+
+#define _BXT_STRINGIFY(x) #x
+#define BXT_TOSTRING(x) _BXT_STRINGIFY(x)
+#define BXT_CONCATENATE(x, y) x##y
+#define BXT_CONCATENATE_DETAIL(x, y) BXT_CONCATENATE(x, y)
```
YaLTeR, line 54: Maybe just BXT_CONCAT()? Also what does "DETAIL" mean?

BunnymodXT/shared.hpp
```diff
@@ -45,3 +45,11 @@ constexpr int kRenderFxTrigger = 241; // DO NOT CHANGE THIS VALUE OR YOU WILL BR
...
+#define BXT_FUNC __func__
```
YaLTeR, line 55: Is this needed? It's the same length as just `__func__`

BunnymodXT/shared.hpp
SmileyAG:
```cpp
#define test RTY
BXT_TOSTRING(BXT_CONCATENATE(QWE, test)) // "QWEtest" string
BXT_TOSTRING(BXT_CONCATENATE_DETAIL(QWE, test)) // "QWERTY" string
```

BunnymodXT/shared.hpp
SmileyAG: You know, for the sake of greater flexibility, I'd rather wrap even this, despite the fact that `__func__` has been a standard since C99 and C++11 and is supported by most compilers.

Like there are `__PRETTY_FUNCTION__` and `__FUNCTION__` for the same purpose, which as far as I know, are not supported by all compilers, but you never know if you might want to replace with them later, so replacing one word in a macro a little faster and easier than replacing `__func__` references in all files with one button.

BunnymodXT/shared.hpp
YaLTeR: Yeah but what does the word "DETAIL" have to do with this?

YaLTeR: I like EXPAND better yeah

PR #535, opened 2024-07-14
minor: triangle_drawing: added template function for drawing trigger polygons

SmileyAG:
For drawing trigger polygons on pre and post 25th anniversary versions.

### DIFF
--- a/BunnymodXT/triangle_drawing.cpp
+++ b/BunnymodXT/triangle_drawing.cpp
@@ -160,6 +160,17 @@ namespace TriangleDrawing
 		return s;
 	}
 
+	template <typename T> void DrawPolygons(triangleapi_s *pTriAPI, const model_t *model, int i)
+	{
+		const T surfs = (T)model->surfaces + model->firstmodelsurface;
+		pTriAPI->Begin(TRI_POLYGON);
+		for (int j = 0; j < surfs[i].polys->numverts; ++j)
+		{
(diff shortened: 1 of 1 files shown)


YaLTeR:
Wait so shouldn't there be two different invocations of the template then? How would it look?

I'm asking because I'm not sure this cast should be done this exact way, but I need to see both invocations to tell for sure.

SmileyAG:
> but I need to see both invocations to tell for sure.

```cpp
if (hw.is_hl25)
  DrawPolygons<msurface_25th_anniversary_hw_t*>(pTriAPI, model, i);
else
  DrawPolygons<msurface_t*>(pTriAPI, model, i);
```

Compilers will simply create two functions in binary, and in the function where triggers are drawn, it will add **CALL** instructions to them (`E8 ?? ?? ?? ??`)
Inside the function where polygons are drawn, there is a cast to the required data type in the form: `(T*)model->surfaces`, so there are clearly no reasons here that point to UB, if this suddenly worries you.

But if your question actually was directed at why I didn’t do it initially, well it was because there is no HL25 support as such yet, although I can comment this piece of code in advance for later uncomment when it comes to it.

YaLTeR:
Then if I understand correctly, that cast inside should be a `reinterpret_cast<T>` instead of `(T)` to prevent some of potential misuse.

PR #536, opened 2024-07-14
optimization: triangle_drawing: don't call GetTrigger(Color/Alpha) functions more than once

SmileyAG:
It is simply pointless to obtain this information more than once and it cost in performance too.


YaLTeR:
Thanks, must've missed it when it was added.

PR #540, opened 2024-07-14
helper_functions: added IsPlayer and Center own implementations to not rely on hooked virtual functions

SmileyAG:
If you want an answer, is it really necessary to overloading the same function first with each individual argument, and then only with `edict_t` as an argument? I'll answer yes, because it solves two issues at once:

- In defense of `edict_t`: you can simply give only the edict as input, and the rest the function will do everything for you, wonderful, isn’t it?
- In defense of the function with each separate argument: this also allows for use in client code, for example, where you can get similar information but not with `edict_t`

Also In case someone wants to merge it if I would inactive for a while: simply rebase to the latest master as soon as https://github.com/YaLTeR/BunnymodXT/pull/531 is merged to solve the build errors

### DIFF
--- a/BunnymodXT/helper_functions.cpp
+++ b/BunnymodXT/helper_functions.cpp
@@ -4,6 +4,30 @@
 
 namespace helper_functions
 {
+	bool IsPlayer(const edict_t *ent)
+	{
+		// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/player.cpp#L2850
+
+		auto &hw = HwDLL::GetInstance();
+		auto &sv = ServerDLL::GetInstance();
+
+		if (strcmp(hw.GetString(ent->v.classname), "player") != 0)
+			return false;
+
+		if (!(ent->v.flags & FL_CLIENT))
+			return false;
+
+		if (sv.pEngfuncs && hw.ppGlobals)
+		{
+			int index = sv.pEngfuncs->pfnIndexOfEdict(ent);
+
+			if ((index < 1) || (index > hw.ppGlobals->maxClients)) // gGlobalVariables.maxClients = svs.maxclients
+				return false;
+		}
+
+		return true;
+	}
+
 	void com_fixslashes(std::string &str)
 	{
 		// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/game_shared/bot/nav_file.cpp#L680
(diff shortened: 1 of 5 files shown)


BunnymodXT/helper_functions.cpp
```diff
@@ -4,6 +4,30 @@
...
+		// https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/player.cpp#L2850
+
+		auto &hw = HwDLL::GetInstance();
+		auto &sv = ServerDLL::GetInstance();
+
+		if (strcmp(hw.GetString(ent->v.classname), "player") != 0)
+			return false;
+
+		if (!(ent->v.flags & FL_CLIENT))
+			return false;
```
YaLTeR, line 21: I'd move this above the strcmp because it's a much cheaper check.

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2623,12 +2623,8 @@ bool ServerDLL::IsPlayerMovingPushable(const entvars_t *pevPushable, const entva
 	if (pevToucher->flags & FL_ONGROUND && pevToucher->groundentity && &pevToucher->groundentity->v == pevPushable)
 		return false;
 
-	/*
-	void *pToucher = pevToucher->pContainingEntity->pvPrivateData;
-	_IsPlayer IsPlayerFunc = *reinterpret_cast<_IsPlayer *>(*reinterpret_cast<uintptr_t *>(pToucher) + offFuncIsPlayer);
-	if (!IsPlayerFunc(pToucher))
+	if (!helper_functions::IsPlayer(pevToucher->pContainingEntity))
```
YaLTeR, line 2600: I'd squash this commit into the previous one

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2690,19 +2690,10 @@ HOOK_DEF_6(ServerDLL, int, __fastcall, CBasePlayer__TakeDamage, void*, thisptr,
...
-
-			Vector playerCenter;
-			Vector inflictorCenter;
-			playerCenterFunc(thisptr, edx, &playerCenter);
-			inflictorCenterFunc(pInflictor, edx, &inflictorCenter);
-			*/
-
-			Vector playerCenter;
-			Vector inflictorCenter;
+			entvars_t *pev = *reinterpret_cast<entvars_t**>(reinterpret_cast<uintptr_t>(thisptr) + 4); // _Smiley: TODO: replace it with GET_PEV macro once it's merged.
```
YaLTeR, line 2693: Same, I'd squash. And now you can GET_PEV too

YaLTeR (commented):
> If you want an answer, is it really necessary to overloading the same function first with each individual argument, and then only with edict_t as an argument?

Makes sense

PR #541, opened 2024-07-14
helper_functions: allow multiple instances on Windows

SmileyAG:
Simply moved code from HwDLL to helper_functions for more convenient management

The only new change is that now it’s handled for Sven Co-op too


YaLTeR:
Uh, maybe I should've checked CI before merging

PR #544, opened 2024-07-14
helper_functions: added IsBSPModel and get_origin_of_entity functions

SmileyAG:
Rebase after #540 is merged

`IsBSPModel` - function from HLSDK to determine if it is brush entity or not

`get_origin_of_entity` - custom function to get origin based on if it is brush entity or not (but this does not work correctly with some entities, e.g. `trigger_`, because those coordinates are also based on the center point, but the `IsBSPModel` from HLSDK does not take them into account, and I don’t want to use hacks yet for that because this is not really necessary for the current code)

### DIFF
--- a/BunnymodXT/modules/HwDLL.cpp
+++ b/BunnymodXT/modules/HwDLL.cpp
@@ -4628,8 +4628,7 @@ void HwDLL::PrintEntity(std::ostringstream &out, int index)
 	if ((!strncmp(classname, "func_door", 9)) || (!strncmp(classname, "func_rotating", 13)) || (!strncmp(classname, "func_train", 10)))
 		out << "; dmg: " << ent->v.dmg;
 
-	Vector origin;
-	HwDLL::GetInstance().GetOriginOfEntity(origin, ent);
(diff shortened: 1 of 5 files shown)


BunnymodXT/modules/HwDLL.cpp
```diff
@@ -4777,22 +4776,6 @@ struct HwDLL::Cmd_BXT_Print_Entities_By_Index
...
 };
 
-void HwDLL::GetOriginOfEntity(Vector& origin, const edict_t* ent)
-{
-	const auto& hw = HwDLL::GetInstance();
-	const char* classname = hw.GetString(ent->v.classname);
-	bool is_trigger = std::strncmp(classname, "trigger_", 8) == 0;
-	bool is_ladder = std::strncmp(classname, "func_ladder", 11) == 0;
-	bool is_friction = std::strncmp(classname, "func_friction", 13) == 0;
-	bool is_water = std::strncmp(classname, "func_water", 10) == 0;
```
YaLTeR, line 4787: This seems relevant at least for teleport to entity?

PR #550, opened 2024-08-12
optimization: initialize game directory only once and assign personal ID to it

SmileyAG:
I didn't have time to check if there were any typos or mixed up IDs, so I'll take a look again after likely today later and give a final comment.

For now, I've requested it for the sake of reviewing the implementation as a whole. Of course, ideally, I'd like to split some of the changes into several pull requests, but then the review and their rebase will take even longer, and you want a release as soon as possible, so you'll have to accept it as it is now. Yeah...

Compared to the old code:

- Gamedir is initialized only once! (including path cleanup and lowercase conversion)

Before this, all these actions occurred with each call of a function somehow related to the game directory, which of course you understand how it can affect performance.

- We now check not the strings of game directories, but instead we make them personal IDs and compare them by them, it is more convenient to work with this, and besides, it should also add to performance.


BunnymodXT/modules/ClientDLL.cpp
```diff
@@ -1280,9 +1202,9 @@ bool ClientDLL::DoesMapNameContain(const char *map)
 	GetMapName(map_name, ARRAYSIZE_HL(map_name));
 
 	if (map_name[0])
-		ConvertToLowerCase(map_name);
+		helper_functions::convert_to_lowercase(map_name);
 
-	return std::strstr(map_name, map);
```
YaLTeR, line 1285: Why? It's called "DoesMapNameContain" not "DoesMapNameStartWith"

BunnymodXT/modules/HwDLL.cpp
```diff
@@ -6860,14 +6866,25 @@ std::string GetGameDir(bool lowercase)
 		}
 	}
 
-	if (lowercase) ? return gamedir_lw; : return gamedir;
+	if (lowercase)
```
YaLTeR, line 6869: These changes sound like they also need to be in the last commit (or vice versa). At the very least to remove clearly broken code from commits

YaLTeR (changes requested):

PR #551, opened 2024-08-12
pre/post: reworked changes in AddToFullPack for correctness and portability

SmileyAG:
Three reasons why this request exists:

- In HLSDK 1.0-based engines and game directories, AddToFullPack is not in the server-side (hl.dll), but in the engine (hw.dll) under the name `SV_AddToFullPack`

So using such a pre/post implementation, we can easily make the same changes in several similar functions, i.e. no need to stupidly copy-paste the code every time

- We return old states only when they have actually been changed, thereby removing the possibility of accidentally overwriting the new state for objects to which we should not even make changes!

- For integer variables that serve to store flags for performing bitwise operations with them, we now return not their entire old state, but only the old state of a specific flag that we changed!

### DIFF
--- a/BunnymodXT/modules/ServerDLL.cpp
+++ b/BunnymodXT/modules/ServerDLL.cpp
@@ -2508,73 +2508,96 @@ void ServerDLL::GetTriggerAlpha(const char *classname, bool inactive, bool addit
 		a = common_alphas[inactive][additive];
 }
 
+void ServerDLL::AddToFullPack_Pre(edict_t *ent, unsigned char **pSet, int &status)
+{
+	if (CVars::bxt_render_far_entities.GetBool()) // https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/dlls/client.cpp#L1114-L1122
+		*pSet = NULL;
+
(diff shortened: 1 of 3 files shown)


BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -2508,73 +2508,96 @@ void ServerDLL::GetTriggerAlpha(const char *classname, bool inactive, bool addit
...
-	ent->v.effects = oldEffects;
-	ent->v.rendermode = oldRendermode;
-	ent->v.rendercolor = oldRenderColor;
-	ent->v.renderamt = oldRenderAmount;
-	ent->v.renderfx = oldRenderFx;
-	ent->v.flags = oldFlags;
-	ent->v.iuser1 = oldIUser1;
-	ent->v.iuser2 = oldIUser2;
+	if (status != BXT_ADDTOFULLPACK_STATE_NO) {
+		ent->v.effects = SET_OR_UNSET_FLAG(oldEffectsNodraw, ent->v.effects, EF_NODRAW);
```
SmileyAG, line 2592: Damn, how did I let this happen and the compiler didn't even complain? Will fix it later simply to:

`SET_OR_UNSET_FLAG(oldEffectsNodraw, ent->v.effects, EF_NODRAW);`

PR #555, opened 2024-09-24
Enable dependabot to update submodules and actions

permanentdaylightt:


YaLTeR:
Hey, sorry for the delay.

I'm not sure we need this? Like, our SPTLib, hlstrafe, taslogger submodules are just other BXT projects that I update in unison with BXT itself (they also require BXT code changes on update, i.e. cannot be updated automatically). Our discord-rpc is an ancient version that will never be updated. Our cereal and rapidjson don't really need updates, they just work, we aren't using them for anything security-sensitive.

Actions are honestly not a big deal to update by hand.

permanentdaylightt:
You can merge if you want

permanentdaylightt:
todos:
make actions upload zips instead of 7z
use own install-boost script (some fixes)

khanghugo:
I went through to look at some changes you made. Beside the obvious that this isn't very important change, there are something to day about your proposal. 7zip is superior, period. GitHub CI artifact upload as ZIP does not retain file permission so your ZIP artifact might be different. The install-boost is a fork from your repository and is not upstream so that means we (ala Yalter) have to vet your repo as well.

permanentdaylightt:
youre right actually, im gonna make some changes to revert it

YaLTeR:
To reiterate, I still do not see why this PR is useful. Definitely not with an unvetted action fork.

permanentdaylightt:
well i made this so there will be no need to update the submodules and the github actions

PR #578, opened 2026-02-11
Added PM_CheckStuck hook to prevent non-native gameplay changes

Tankfird:
The PM_PlayerMove hook was calling PM_CheckStuck() for bxt_fire_on_stuck check. The issue with this is the function tries to unstick the player every few frames using `static int rgStuckLast[MAX_CLIENTS][2];` to keep track of the last offset to use. The call in the hook would sometimes cause it to take the path
```
i = PM_GetRandomStuckOffsets(pmove->player_index, pmove->server, offset);
VectorAdd(base, offset, test);
if ( ( hitent = pmove->PM_TestPlayerPosition ( test, NULL ) ) == -1 )
{
	//Con_DPrintf("Nudged\n");

	PM_ResetStuckOffsets( pmove->player_index, pmove->server );

	if (i >= 27)
		VectorCopy ( test, pmove->origin );

	return 0;
}
```
```
int PM_GetRandomStuckOffsets(int nIndex, int server, vec3_t offset)
{
 // Last time we did a full
	int idx;
	idx = rgStuckLast[nIndex][server]++;

	VectorCopy(rgv3tStuckTable[idx % 54], offset);

	return (idx % 54);
}

void PM_ResetStuckOffsets(int nIndex, int server)
{
	rgStuckLast[nIndex][server] = 0;
}
```
Which resets the rgStuckLast to 0 and normally `return 0;` would cause normal player movement code to run, this is skipped in the hook.

Ingame you can use `bxt_ch_set_pos_offset 0 0 -0.1` with `bxt_hud_velocity 1` while moving on the ground, normal stuck behavior you'll see velocity quickly get cleared out with just gravity in zvel. Currently with bxt loaded it might never reach that and you'll maintain velocity while stuck. This may have affected some op4 speedruns as we use the rope to get stuck in the ground with velocity to then object boost for speed. Also may have allowed some other stuck tricks to work (tripmines, snarks, etc.), might desync some TAS.


SmileyAG:
I am so sorry for this unintentional old bug and hate myself for it a bit now.

If you want to know why it happened, as far as I can remember, story was something like this:

Back to late 2022, the amateur brain seriously believed in that MSVC/GCC compilers would optimize his code so well that the variable will be set directly at the moment of next check (e.g. `&& (stuck_cur_frame = ORIG_PM_CheckStuck()) &&` instead of `&& stuck_cur_frame &&`), since brain thought that the starting check as `if (!CVars::bxt_fire_on_stuck.IsEmpty()` wouldn't allow to execute further and therefore to not afraid that this function will be called earlier out of place (hard to imagine how to came to this conclusion since this barely makes a glimpse of sense or to be fair rather not at all - but amateur didn't think like that previously).

This could be ended if that brain was ever thought at least to verify in practice his hypothetical theories with a simple tests in Godbolt or locally as testing grounds - no instead we thought that request it right away to upstream right away would be great.

Each change must be further checked additionally by one of the contributors in order to completely exclude such possibility before release. Extra pair of eyes and different mind thinking never hurts. 

Not a fan of modern trend to entrusting this tasks to AI agents at all with a massive slop reports of beating around the bush, considering that we still have a couple of active old/new contributors with experience who I believe wouldn't refuse to spare only a few minutes of free time for reviewing requests and everyone can review the pull requests of each other on weekends.
AI is a productive tool in the experienced hands for sure, but I'm totally against of it being used by rookies who don't have a idea of what they building with it - it's fine if used for local projects or as a PoC, but not when it's blindly merged to the upstream - especially it hurts when it comes to project you are so passionate to it.

SmileyAG:
Close to the actual topic, @Tankfird confirmed that these groundentity skips can be reproduced without the BXT injected – so it's not as damaging much, but still this bug gives a some benefit.

> Tankfird — 05:30
> Yeah I retested them and it all worked. It's mostly shallow stucks that are affected and because it retries each checkstuck it can take some time to for it to reset the playerstate. Most likely it will only affect the tripmine stucks when you're just at the edge in a certain way that it's an early offset check
﻿
> Tankfird — 05:48
> I think it likely affects the op4 strats more, rope skip + water tripmine
> because you're stuck longer

I think the runs on current BXT version should be still accepted whatever you like it or not, otherwise it would be so cruel to reject the world record from Tankfird and others runs due to a bug that is not depends from the runners at all.
But releasing a new hotfix version would be a great too to not to encourage this behavior and stops to make it as the norm.

So this fix should not affect much at least for HL1 runs according to @Tankfird, but if runners would came out later to conclusion in discussions that this makes is too inconsistent of glitch reproduce or significantly affect the time on execution up to seconds instead of a few frames, then sure we may reconsider the decision and release another a hotfix release with cvar, which will default to the traditional game behavior - but allow to toggle for the wrong old behavior.

To be honest, this is not even the first time such a case arises, like `bxt_autojump` code since its existence has flaws that did not take into account the call **CBasePlayer::Jump** and therefore on conveyor belts as from the RP chapter in HL1 you gain a less speed than you suppose to, `+bxt_tas_autojump` is more legitimate and mod-support friendly alternative in this regard.

SmileyAG (approved):

BunnymodXT/modules/ServerDLL.cpp
```diff
@@ -1983,6 +1969,23 @@ HOOK_DEF_1(ServerDLL, void, __cdecl, PM_PlayerMove, qboolean, server)
 	CustomHud::UpdatePlayerInfo(velocity, origin);
 }
 
+HOOK_DEF_0(ServerDLL, int, __cdecl, PM_CheckStuck) 
+{
+	HwDLL &hwDLL = HwDLL::GetInstance();
+	bool stuck_cur_frame = false;
+	static bool not_stuck_prev_frame = false;
+	stuck_cur_frame = ORIG_PM_CheckStuck();
```
SmileyAG, line 1977: Thought of requesting to replace with `auto ret = ORIG_PM_CheckStuck();` to keep consistency with rest of codebase, but this is cosmetic edit after all so who cares truly `¯\_(ツ)_/¯`

Tankfird: yeah feel free to change the style. I'm not terribly familiar with everything and also I think I tried to keep everything relatively the same from the old stuck check

YaLTeR:
LGTM

Muty71:
@SmileyAG Creating a toggle to emulate a bug that causes non-native gameplay with the purpose of using it for speedruns goes against the spirit of why BXT was created in the first place, in my opinion. Very last resort territory.

SmileyAG:
@Muty71 @Tankfird Responding to a previous message, I remember you earlier raising concerns that this might potentially have affected the route at OpFor.

So I decided personally to contact javac (current world record holder in the OpFor scripted category) and ask javac to test the new hotfix update in OpFor, so javac replied to me later and informed that with the new hotfix update there were no changes at all in the consistency and timings of rope skip / tripmine water clipping, so I think we can breathe out and don't comeback to this topic ever again (oh thank goodness). 

So the current hotfix behavior will keep remain since as it was supposed to be in original game after all.

SmileyAG:
In addition, I decided to manually dig through the current codebase of server and engine modules to see:
- Unintentional use of functions or set variables outside of conditions that change the game behavior of the game (this what exactly happened in this case)
- The use of functions whose logic carries much more than only reading and returning the state as it may looks in first glance and therefore affect the behavior of the game as the end of result.

I haven't been able to find anything suspicious so personally I don't see reason to concern for now, but if something would comes up in future, please report it and open the PR with the proposed fix if possible. 
Thank you @Tankfird for your attention to bring this fix!

